PR #1, opened 2017-08-07
Add backslash lost with text changes

AloisMahdal:
A backslash was lost during language-related changes, so solution text
is printed to stdout instead of solution.txt file.

This commit puts the backslash back.


bocekm (dismissed):

bocekm:
--test--

bocekm:
--test--

bocekm:
@AloisMahdal, please rebase with el6toel7 so the PR can be built in Copr. Another option is to drop temporarily the requirement of successful Copr build before merging :)

AloisMahdal:
> Another option is to drop temporarily the requirement of successful Copr build before merging...

That would be blasphemy!  Let's not spoil this young repo... :-)

Rebased...

pirat89 (approved):
Good catch! Thanks. It's ok now.

Issue #2, opened 2017-08-08
Do not use removed API functions/variable


bocekm:
See https://github.com/upgrades-migrations/preupgrade-assistant/pull/291 for details.

PR #10, opened 2017-09-13
use current API for modules from PA v2.4.5

pirat89:
The new API for modules has been changed:
 -- removed $RESULT_FAILED (only $RESULT_FAIL remains)
 -- removed check_root that is obsoleted

Fixes: #2


jmazanek (approved):
Tested. Haven't seen any issues

PR #11, opened 2017-09-13
 networking/samba

jmazanek:
- message printed regarding the deprecation of SMB1 protocol

### DIFF
--- a/RHEL6_7/networking/samba/check
+++ b/RHEL6_7/networking/samba/check
@@ -1,0 +1,10 @@
+#!/bin/bash
+. /usr/share/preupgrade/common.sh
+
+#END GENERATED SECTION
+
+
+
+
+log_slight_risk "SMB1 protocol is deprecated and it is replaced by SMB2."
+exit $RESULT_FAIL
--- a/RHEL6_7/networking/samba/module.ini
+++ b/RHEL6_7/networking/samba/module.ini
@@ -1,0 +1,7 @@
+[preupgrade]
+content_title = SMB 
+content_description = SMB1 protocol has been deprecated. The module informs the user about the risks
+author = Jakub Mazanek <jmazanek@redhat.com>
+check_script = check
+check_applies_to = samba-common
+solution = solution.txt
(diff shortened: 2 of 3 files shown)


RHEL6_7/networking/samba/check
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+. /usr/share/preupgrade/common.sh
+
+#END GENERATED SECTION
+
+
+SOLUTION_FILE="./solution.txt"
```
pirat89: Why do you create this variable for one use in echo? It is unneeded.

RHEL6_7/networking/samba/check
```diff
@@ -0,0 +1,12 @@
...
+
+echo "SMB1 protocol is deprecated as it is replaced by SMB2 protocol. It is now considered insecure, due to it's vulnerability to man-in-the-middle attacks. Support for the SMB2 protocol will not be backported to Red Hat Enterprise Linux 6 version of samba\* packages. Therefore, for security and compatibility reasons, we strongly recommend to the user to upgrade whole Red Hat Enterprise Linux based server - clients network at once." > $SOLUTION_FILE
```
pirat89: When the message is persistent and there is not any other check than the samba-common package available on system, the message shoud be part of the solution.txt file.

RHEL6_7/networking/samba/check
```diff
@@ -0,0 +1,12 @@
...
+
+log_high_risk "SMB1 protocol is deprecated."
+exit $RESULT_FAIL
```
pirat89: High risk is not needed here as the action is required on the target machine. Use rather the `log_medium_risk` instead.

pirat89 (changes requested):

jmazanek:
Module has been edited in accordance with requested changes.  Message text has been already reviewed by jhornice. Please merge

pirat89 (dismissed):
It seems OK. I'll keep it opened as changes in text are expected yet and it can be merge everytime before the next release.

jmazanek:
The link to the article discussing in detail implications of the change for the interworking with Windows based networks was added to the module

pirat89 (approved):

PR #12, opened 2017-09-13
Port prep_symlinks() and friends as module set init script

AloisMahdal:
Note that this changes the exact format of the resulting error message
as preupgrade-assistant will present it in a more generic way as "module
init script failed; here's stderr", while the original version would
just show what's now in 'stderr'.

### DIFF
--- a/RHEL6_7/init
+++ b/RHEL6_7/init
@@ -1,0 +1,137 @@
+#!/usr/bin/python
+
+from __future__ import unicode_literals
+import os
+import shutil
+import sys
+from distutils import dir_util
+from preupg.utils import SystemIdentification, FileHelper
+from preupg import settings
+
+
+class ModuleInit(object):
+
+    """Class handles with common log files"""
+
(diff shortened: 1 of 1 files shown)


pirat89 (approved):
Modules work correctly with this change and I haven't found any issue from the side of this script. Just the `server_variant` and `server_variant_files` variables would be renamed to something more common like `system_*`, because there are 4 variants of RHEL 6 system: *Client*, *ComputeNode*, *Workstation*, *Server*. But it is just cosmetic change. @AloisMahdal What do you think Lojzo? I am keeping it opened for now in case you want to change it.

AloisMahdal:
I do want to change it, but best practice is to keep PR focused and simple, which is the sole reason why I did not.  (I actually did rename `server_variant` to `variant` but then reverted it in order to make the PR more reviewable.)

So yes, script should be refactored but I suggest opening new PR for that.

pirat89:
Thanks for response, so I am merging it to the master.

PR #14, opened 2017-09-17
openssh/sysconfig: update regexp, text and use log_info

pirat89:
Use the log_info function instead of log_slight_risk as risks are
not allowed when the FIXED result is expected. Additionaly accept
whitespaces before the 'export' command, previously script could
end with PASS even when the 'export' command has been used but no
one occured on beggining of line.

Added "-q" option for the grep in the condition to supress leaked
output. Texts has been modified little too.

### DIFF
--- a/RHEL6_7/services/openssh/sysconfig/check
+++ b/RHEL6_7/services/openssh/sysconfig/check
@@ -1,32 +1,33 @@
 #!/bin/bash
 
 . /usr/share/preupgrade/common.sh
-check_applies_to "openssh-server"
 
 #END GENERATED SECTION
 
 
+SYSCONFIG_FILE="/etc/sysconfig/sshd"
+CLEAN_SYSCONFIG_FILE="$VALUE_TMP_PREUPGRADE/cleanconf/$SYSCONFIG_FILE"
 
+mkdir -p $(dirname $CLEAN_SYSCONFIG_FILE)
+cp -a $SYSCONFIG_FILE $CLEAN_SYSCONFIG_FILE
 
-SYSCONFIG_FILE=/etc/sysconfig/sshd
+if grep -q "^[[:space:]]*export[[:space]]" $SYSCONFIG_FILE; then
+    msg="The $SYSCONFIG_FILE file will not be a shell script in Red Hat"
+    msg+=" Enterprise Linux 7 anymore, so all 'export VARIABLE=VALUE' have to"
+    msg+=" be changed to 'VARIABLE=VALUE'."
+    echo -e "$msg" >> solution.txt
 
-mkdir -p $VALUE_TMP_PREUPGRADE/cleanconf/$(dirname $SYSCONFIG_FILE)
-cp $SYSCONFIG_FILE $VALUE_TMP_PREUPGRADE/cleanconf/$SYSCONFIG_FILE
+    log_info "The 'export' commands will be removed from the $SYSCONFIG_FILE file."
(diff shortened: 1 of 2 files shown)


jmazanek (approved):

AloisMahdal:
`qa_nak`: Please understand that in order to test, we need to know how *behavior* should change under what circumstances.  Also it's hard to work with PRs that do multiple irrelevant things at once (you know this is a bad practice).

pirat89:
You know that I am sometimes lazy.. but only sometimes :-) Basically it's like rewrite of module, just you will not see "so big" changes in behaviour. I will write specification

pirat89:
Meh ?! I found there small issue :/ log_info missing. I will create new PR that will fix it.
Edit: See PR #27 .

pirat89:
The module specification:
===================

short story long:
-----------------
  The module is applid when openssh-server is installed; In that case,
  always back up [0] to [1]. When export command is used inside [0], print
  solution about changes in RHEL 7 and try to remove export command automatically
  from [1]. When "sed" utility sucesses, log info and set result fixed. In case
  that sed fails, require manual action from user (high risk + exit_fail).

long story short:
-----------------
```
INPUT FCS  || BEHAVIORS
INST | EXP || BAK | FIX | RESULT
=====|=====||=====|=====|===============
 no  | -   || -   | -   | not_applicable
 yes | no  || yes | -   | pass
 yes | yes || yes | no  | needs_action (untestable probably)
 yes | yes || yes | yes | fixed
```
legend:
  - INST: openssh-server is installed 
  - EXP:  The config file [0] uses export command to set and export environment
        variables for "relative" utilities
  - BAK:  back up [0] to [1]
  - FIX:  the "export" command has been removed from [1] (sed returns 0)


[0] /etc/sysconfig/sshd
[1] /root/preupgrade/cleanconf/etc/sysconfig/sshd

**update by @aloismahdal**: just fixed the table to separate input vs. behavior.

AloisMahdal:
Now that's the spirit! :-)

AloisMahdal:
`qa_ack`: Writing automated test according to specification above looks plausible.

pirat89:
I have doubts now about correctness of this (my) solution as according to the solution text, it seems that you can write whatever inside the config file; like:
```bash
fname() { export somevar=someval; }
fname
```
And it would still work. I didn't test it but in case it works, the testing regexp is not so good.

AloisMahdal:
Well if the "config" file is a shell script then without executing it you just *don't have* any way of knowing what it does (let alone by a regex).  There are virtually infinite ways to export a variable from a shell script. (Including sourcing of different files and meta-programming, of which there are even "more infinite" ways.)

AloisMahdal:
(OTOH, you *can* find it out it by execution, given there are safe assumptions about how exactly it should be executed.)

AloisMahdal:
Tested with released version:

    grep: Unmatched [ or [^

Also please see my comments above.

pirat89:
@AloisMahdal: You are right. But in that case, about logic of the module at all, because in that case, it
would always ended as informational, to provide info about that change between 6 and 7 - where it can't be script anymore. And I am interested about use-case for something like that. Executable "config" file is for me always wrong idea.

pirat89:
Ha. I see. There is missing semicolon. Now I am curious why I didn't see that earlier. :/
Probably I used different version of modules for testing as I see. Because with this nobody can see result *fixed*.

PR #15, opened 2017-09-18
Do not reference old API - RESULT_FAILED

AloisMahdal:
PR #10 missed this.


pirat89:
Rather than that, I would prefer completely remove these comments from modules as they are pretty useless.

AloisMahdal:
> Rather than that, I would prefer completely remove these comments from modules as they are pretty useless.

Fair enough.  I'm preparing new PR for that.

PR #16, opened 2017-09-18
Remove obsolete and possibly confusing comments

AloisMahdal:
These comments come from template and are not necessary in the modules
(actually they are rather confusing).


pirat89 (approved):
Yes, my pleasure to merge it :-)

Issue #17, opened 2017-09-18
init script: catch traceback when env variable not set (KeyError)


bocekm:
Print some sensible output to STDERR, like _"Environment variable \<var\> needs to be set."_

pirat89:
Current output:
```bash
ModuleSetInitError: Error: Module set init script reported error:
    status: 1
    stderr: Traceback (most recent call last):
    stderr:   File "/usr/share/preupgrade/RHEL6_7/init", line 135, in <module>
    stderr:     scenario = os.environ['PREUPGM_INIT_SCENARIOs'],
    stderr:   File "/usr/lib64/python2.6/UserDict.py", line 22, in __getitem__
    stderr:     raise KeyError(key)
    stderr: KeyError: u'PREUPGM_INIT_SCENARIOs'

```

AloisMahdal:
`u'PREUPGM_INIT_SCENARIOs'`? I do *not* like the `s` at the end... :)

AloisMahdal:
Anyway, is this really necessary?

I mean, user (ie. module developer) does not have any control of the environment; only one place that does is `Application:run_init()`. So if the variable is missing, then only place that needs fixing is that very method.

IOW, this is not user-facing situation; I don't think this can normally happen.

AloisMahdal:
> u'PREUPGM_INIT_SCENARIOs'? I do not like the s at the end... :)

OK, I see now, you just hacked it just to see the traceback... Nice :)

bocekm:
It's nothing critical, just usual "assert" that the required input parameters are available and preventing user from seeing tracebacks. You're right that the user cannot control this, but we shouldn't just expect it won't happen. I guess the error message may be prepended by "Internal error". But it's up to @pirat89, if you close the issue I won't be mad :)

AloisMahdal:
OK., I see, it's true that the message seems to "blame" `init` file, who can't fix the problem.  IOW if anything should be fixed, then it's in run_init.

But that takes us back: many things in `Application` already do depend on these values, so why `run_init()` in particular should be the one that does the checking?  (Actually only role of `run_init()` is that it translates few mandatory properties of `Application()` instance to environment variables.

The way @pirat89 reproduced the traceback above is by injecting typo in `init`. If people make typo, they can immediately see the offending key in the traceback--it's simple thing, wrapping it in other error catching/translation code is IMO not worth it.

AloisMahdal:
[...]

PR #18, opened 2017-09-19
init: added check of env settings

pirat89:
Print error message and exit with exit code 1 when any of required
environment variables is not set.

Resolves: #17

### DIFF
--- a/RHEL6_7/init
+++ b/RHEL6_7/init
@@ -128,10 +128,14 @@ class ModuleInit(object):
 
 if __name__ == "__main__":
 
-    i = ModuleInit(
-        assessment_dir = os.environ['PREUPGM_INIT_ASSESSMENT_DIR'],
-        dst_arch = os.environ['PREUPGM_INIT_DST_ARCH'],
-        contents = os.environ['PREUPGM_INIT_CONTENTS'],
-        scenario = os.environ['PREUPGM_INIT_SCENARIO'],
-    )
+    try:
+        i = ModuleInit(
+            assessment_dir = os.environ['PREUPGM_INIT_ASSESSMENT_DIR'],
+            dst_arch = os.environ['PREUPGM_INIT_DST_ARCH'],
+            contents = os.environ['PREUPGM_INIT_CONTENTS'],
+            scenario = os.environ['PREUPGM_INIT_SCENARIO'],
+        )
+    except KeyError, e:
+        sys.stderr.write("The '%s' environment variable is not set." % str(e))
+        sys.exit(1)
     i.prep_symlinks()


RHEL6_7/init
```diff
@@ -128,10 +128,14 @@ class ModuleInit(object):
...
-        scenario = os.environ['PREUPGM_INIT_SCENARIO'],
-    )
+    try:
+        i = ModuleInit(
+            assessment_dir = os.environ['PREUPGM_INIT_ASSESSMENT_DIR'],
+            dst_arch = os.environ['PREUPGM_INIT_DST_ARCH'],
+            contents = os.environ['PREUPGM_INIT_CONTENTS'],
+            scenario = os.environ['PREUPGM_INIT_SCENARIO'],
+        )
+    except KeyError, e:
```
AloisMahdal, line 138: This will catch any KeyErrors in `ModuleInit.__init__`.

My suggestion would be to move the try/catch inside the function.  You could achieve that by removing all arguments from the initializer and reading os.environ in there (or passing it as sole argument, eg. `i = ModuleInit(env=os.environ)`.

pirat89: I don't like that solution too much, because it means that you have to call the "exit" function from the `__init__` method and that seems ugly to me. Currently the try/catch is OK according to code inside `__init__` but it doesn't have to be true in future, so I will fix it.

bocekm: I agree with @AloisMahdal and his proposal seems good. Anyway, here's another option, that would get rid of the "ugliness":
```python
if __name__ == "__main__":

    try:
        init_env_vars = {
            'assessment_dir': os.environ['PREUPGM_INIT_ASSESSMENT_DIR'],
            'dst_arch': os.environ['PREUPGM_INIT_DST_ARCH'],
            'contents': os.environ['PREUPGM_INIT_CONTENTS'],
            'scenario': os.environ['PREUPGM_INIT_SCENARIO'],
        }
    except KeyError, e:
        sys.stderr.write("The '%s' environment variable is not set." % str(e))
        sys.exit(1)

    i = ModuleInit(init_env_vars)
```

AloisMahdal: Sorry for the confusion, but I thought this through and nothing seems to really work. and IMO that means this is a bad idea in the first place.  You can't do any useful and maintainable error checking here, because there's no way for lower level (`init`) to babysit upper level (`run_init()`).

IOW, what I said in https://github.com/upgrades-migrations/preupgrade-assistant/pull/311#discussion_r140375004 apples also to this situation.

pirat89:
I am closing the PR as it is more like nice to have then anything else and solution is not nice.

PR #19, opened 2017-09-21
networking/dovecot

jmazanek:
- If first_valid_uid is not explicitly set 
  - the postupgrade script sets it to 500 in order not to brake user's
    service after the upgrade
  - message is printed recommending user to increase the value above 1000
    after upgrade for security reasons
  - needs_inspection
- If it is explicitly set -lt 1000
  - message is printed recommending user to increase the value above 1000
    after upgrade for security reasons
  - needs_inspection
- If it is explicitly set -ge 1000
  - message is printed informing user that the first_valid_uid is
    compatible with RHEL 7
  - pass
- if dovecot configuration is invalid on the source system
  - the need_action is returned along with a message for a user

### DIFF
--- a/RHEL6_7/networking/dovecot/check
+++ b/RHEL6_7/networking/dovecot/check
@@ -9,9 +9,44 @@
 
 # Copy your config file from RHEL6 (in case of scenario RHEL6_7)
 # to Temporary Directory
-CONFIG_FILE="/etc/dovecot/"
-uid_val="$(awk '/^first_valid_uid/ { print $3}' /etc/dovecot/conf.d/10-mail.conf)"
-cp --parents -ar $CONFIG_FILE $VALUE_TMP_PREUPGRADE/dirtyconf
+CONFIG_DIR="/etc/dovecot/"
+
+# Exit if the configuration is invalid
+validate ( ) {
+read conf
+err_msg=$(echo "$conf" | grep -o "\/.*[[:digit:]]")
+if [ "$err_msg" != "" ];then
+    log_high_risk "Error in $err_msg."
+return 1
+fi
+}
+
+doveconf 2>&1- | validate
+
+if [ $? -eq 1 ];then
+    log_high_risk "The Dovecot configuration is invalid. Fix it or remove the dovecot package before proceeding with the upgrade."
+    exit $RESULT_FAIL
+fi
+
+#Get the values of first_valid_uid
+uid_val=$(doveconf -a | awk -F' = ' '/^first_valid_uid/ { print $2}')
+explicit_uid_val=$(doveconf -aN | awk -F' = ' '/^first_valid_uid/ { print $2}')
+
+#Get the list of all possible places, where the first_valid_uid may be defined.
+#List is compiled in line by line manner where the main configuration file
+#continues to be read after each crawl into the included configuration directory.
+declare -a included_files
+while read line; do
+    included_files=("${included_files[@]}" $(echo "$line" | awk '/^\!include(_try)?\>/ {print $2}' ))
+    if echo "$line" | grep -q "^first_valid_uid" ;then
+        included_files=( "${included_files[@]}" dovecot.conf )
+    fi
+done < /etc/dovecot/dovecot.conf
+loaded_conf="$PWD/loaded.conf"
+
+cat /dev/null > "$loaded_conf"
+
+cp --parents -ar $CONFIG_DIR $VALUE_TMP_PREUPGRADE/dirtyconf
 
 #workaround to openscap buggy missing PATH
 export PATH=$PATH:/usr/bin
@@ -53,12 +88,12 @@ then
     rm -f $TMPF1 $TMPF2
   fi
 else
-  log_error "Cannot use doveconf to parse the configuration files."
+  log_high_risk "The Dovecot configuration is invalid. Fix it or remove the dovecot package before proceeding with the upgrade."
   rm -f $TMPF1 $TMPF2
   exit $RESULT_FAIL
 fi
 
-log_info "The configuration files from $CONFIG_FILE will be fixed by the postupgrade script."
+log_info "The configuration files from $CONFIG_DIR will be fixed by the postupgrade script."
 log_slight_risk "In some corner cases your configuration might not be migrated automatically."
 echo "\
 The doveconf tool should be able to migrate your configuration, but there is a slight risk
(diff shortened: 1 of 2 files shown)


jmazanek:
Testing notes:

When testing the module there are certain dovecot configuration pitfalls. Part of the dovecot package is a doveconf utilility, that checks the validity of the dovecot configuration files and returns error message if they don't conform to the valid settings.
SSL protocol  is by default allowed, but it is not by default defined in /etc/dovecot/dovecot.conf file,  therefore if you are testing the use case with no additional configuration files included, you have to provide the following directives inn the main configuration file:

ssl_cert = </etc/pki/dovecot/certs/dovecot.pem
ssl_key = </etc/pki/dovecot/private/dovecot.pem

If only dovecot package is installed , directives that manage the sieve functionality have to be commented out in /etc/dovecot/conf.d/20-managesieve.conf ( or an alternative configuration file) if that is included.  Otherwise you have to install the additional package dovecot-pigeonhole.
Since not conforming to the above mentioned settings causes the dovecot service failure, it is assumed, that if dovecot is actively used on the source system, the above conditions are met.
Otherwise module exits with needs_inspection status and prints the dovecot error message and doesn't proceed with a analysis.

RHEL6_7/networking/dovecot/check
```diff
@@ -53,7 +62,7 @@ then
     rm -f $TMPF1 $TMPF2
   fi
 else
-  log_error "Cannot use doveconf to parse the configuration files."
+  log_high_risk "Dovecot configuration is invalid. Fix it or remove the dovecot package before proceeding with a upgrade"
```
jhornice: "The Dovecot configuration is invalid. Fix it or remove the dovecot package before proceeding with the upgrade"

RHEL6_7/networking/dovecot/check
```diff
@@ -69,13 +78,48 @@ file manually on the new system.
...
+fi
+last_loaded=$(cat "$loaded_conf")
+last_loaded="$VALUE_TMP_PREUPGRADE/dirtyconf/$CONFIG_FILE/$last_loaded"
+export uid_val 
+export last_loaded
+
+if [[ "$explicit_uid_val" == "" ]];then
+    #Change the explcitly stated default value on the target system to the default value on the source system
+    perl -pi -e '$_ .= qq(first_valid_uid = $ENV{uid_val}\n) if /^#first_valid_uid = 500/' "$VALUE_TMP_PREUPGRADE/dirtyconf/$CONFIG_FILE/conf.d/10-mail.conf"
+    log_medium_risk "You are using first_valid_id in the reserved range"
```
jhornice: "You are using first_valid_id in the reserved range."

RHEL6_7/networking/dovecot/check
```diff
@@ -69,13 +78,48 @@ file manually on the new system.
...
+    echo "You are using default first_valid_id of $uid_val for dovecot users, which is in the uid range reserved for system users on Red Hat Enterprise Linux 7. This is not a secure configuration. We strongly recommend you to migrate your regular users to uids above 1000 and increase the value of first_valid_id directive above 1000 on the target system" >> $SOLUTION_FILE
```
jhornice: "You are using the default first_valid_id of $uid_val for Dovecot users, which is in the UID range reserved for the system users in Red Hat Enterprise Linux 7. This is not a secure configuration. Migrate your regular users to UIDs to 1000 and above, and increase the value of the first_valid_id directive to 1000 or above on the target system."

RHEL6_7/networking/dovecot/check
```diff
@@ -69,13 +78,48 @@ file manually on the new system.
...
+    exit_fail
+elif [ "$explicit_uid_val" -lt 1000 ];then
+    log_medium_risk "You are using first_valid_id in the reserved range"
+    echo "You have configured first_valid_id of $explicit_uid_val for dovecot users in $last_loaded, which is in the uid range reserved for system users on Red Hat Enterprise Linux 7. This is not a secure configuration. We strongly recommend you to migrate your regular users to uids above 1000 and increase the value of first_valid_id directive above 1000 on the target system" >> $SOLUTION_FILE
```
jhornice: Stejne jako v predchozim komentari

RHEL6_7/networking/dovecot/check
```diff
@@ -69,13 +78,48 @@ file manually on the new system.
...
+    exit_fail
+
+else
+    log_info "Directive first_valid_uid in your dovecot configuration is compatible with Red Hat Enterprise Linux 7"
```
jhornice: "The first_valid_uid directive in your Dovecot configuration is compatible with Red Hat Enterprise Linux 7."

jhornice (dismissed):
See the comments

AloisMahdal:
qa_ack: Follow specifications in original description and further notes from @jmazanek.

Also, I have an assessment test in progress in downstream test suite, so let me know if you want to take care of this.

jmazanek:
Just to clarify one more thing for testing purposes:
Dovecot sources the main configuration files first, and then the list of included file is alphabetical order. If any of the directives is defined in multiple configuration files, the value from the last sourced file masks any previous value.

pirat89:
Note: rebase and squash of two commits together.

RHEL6_7/networking/dovecot/check
```diff
@@ -10,7 +10,16 @@
...
 CONFIG_FILE="/etc/dovecot/"
-uid_val="$(awk '/^first_valid_uid/ { print $3}' /etc/dovecot/conf.d/10-mail.conf)"
+
+#Get the values of first_valid_uid
+uid_val=$(doveconf -a | awk -F' = ' '/^first_valid_uid/ { print $2}')
+explicit_uid_val=$(doveconf -aN | awk -F' = ' '/^first_valid_uid/ { print $2}')
+
+#Get the list of the included config files in the main config file
+declare -a included_files=$(awk '/^\!include(_try)?\>/ {print $2}' /etc/dovecot/dovecot.conf)
+loaded_conf="/usr/share/preupgrade/RHEL6_7/networking/dovecot/loaded.conf"
```
pirat89: We can't create new files inside that place. You should use: 
 - `$VALUE_TMP_PREUPGRADE/RHEL6_7/networking/dovecot/loaded.conf`
- `$PWD/loaded.conf` - this is better in this case

RHEL6_7/networking/dovecot/check
```diff
@@ -69,13 +78,48 @@ file manually on the new system.
...
+cd /etc/dovecot/
+
+#Get the the config, that actually provides the first_valid_uid directive
+if [[ "${included_files[@]}" != "" ]];then
+    printf '%s\n' ${included_files[@]} | while IFS= read -r included_file 
+    do
+        if [ -e "$included_file" ];then
+           if grep -q "^first_valid_uid" "$included_file" ;then
+               included_file="$included_file"
+           echo "$included_file" > "$loaded_conf"
```
pirat89: indeentation.

RHEL6_7/networking/dovecot/check
```diff
@@ -69,13 +78,48 @@ file manually on the new system.
...
-exit $RESULT_FIXED
+cd /etc/dovecot/
+
+#Get the the config, that actually provides the first_valid_uid directive
+if [[ "${included_files[@]}" != "" ]];then
+    printf '%s\n' ${included_files[@]} | while IFS= read -r included_file 
+    do
+        if [ -e "$included_file" ];then
+           if grep -q "^first_valid_uid" "$included_file" ;then
+               included_file="$included_file"
```
pirat89: this is pointless in that case

RHEL6_7/networking/dovecot/check
```diff
@@ -69,13 +78,48 @@ file manually on the new system.
...
-if ! [[ -z "$uid_val" ]] && [ "$uid_val" -lt 1000 ];then
-  export uid_val ; perl -pi -e '$_ .= qq(sed -i -e "s/first_valid_uid = 1000/first_valid_uid = $ENV{"uid_val"}/g" /etc/dovecot/conf.d/10-mail.conf\n) if  /restorecon -R $CONFIG_FILE/' "$PREF"/dovecot_postupgrade.sh
-fi
 chmod +x $PREF/dovecot_postupgrade.sh
 
 rm -f $TMPF1 $TMPF2
-exit $RESULT_FIXED
+cd /etc/dovecot/
+
+#Get the the config, that actually provides the first_valid_uid directive
```
pirat89, line 112: Actually this is not true, there can be multiple includes and the value can be set even inside main configuration file, when includes are used. And even it can be included between includes mupltiple times - ok it doesn't make sense so much, but user can do that. Solution discussed in person.

RHEL6_7/networking/dovecot/check
```diff
@@ -69,13 +78,48 @@ file manually on the new system.
...
+    fi
+fi
+last_loaded=$(cat "$loaded_conf")
+last_loaded="$VALUE_TMP_PREUPGRADE/dirtyconf/$CONFIG_FILE/$last_loaded"
+export uid_val 
+export last_loaded
+
+if [[ "$explicit_uid_val" == "" ]];then
+    #Change the explcitly stated default value on the target system to the default value on the source system
+    perl -pi -e '$_ .= qq(first_valid_uid = $ENV{uid_val}\n) if /^#first_valid_uid = 500/' "$VALUE_TMP_PREUPGRADE/dirtyconf/$CONFIG_FILE/conf.d/10-mail.conf"
```
pirat89: Issues:
 - what if the line `first_valid_uid = 500` doesn't exist inside the file?
 - what if the file does not exist?

pirat89 (dismissed):
Discussed in person; see comments below

@AloisMahdal : please look at comments and include these cases in your tests if you can.

jmazanek:
- Previously the module incorrectly assessed
   the place from where the dovecot application
   actually sources the first_valid_uid directive.

  - With this fix the module correctly determines
    the directive source

  - If necessary, directive is appended at the end
    of the default configuration file.

jmazanek:
Testing note:
As pointed out in the review by @pirat89, the directives are sourced line by line from the main configuration file.
This also applies to sourcing files specified in the include statements. 
Therefore if there are multiple include statements on the multiple lines or if the directive is specified after the include statement, it rewrites any previous declaration.

jmazanek:
Output texts and comments edited. @jhornice please check them.

pirat89 (dismissed):
Works as expected with limitation (I guess) on recursive use of "includes", when the file included "transitively" is not checked whether sets first_valid_uid value or not.

AloisMahdal:
Tested with new test and build `0.7.0-0.201710271227Z.PR19`.  Found two issues:

 *  If dovecot config is invalid (syntax error injection), the error from doveconf is let out
    to stderr.  Errors should be always reported using risk log or message log.

 *  There is a typo in one of risk messages: `first_valid_id` instead of `first_valid_uid`.

jmazanek:
With the new build Error messages from doveconf are reformulated to use the risk log format. Typos are fixed.

AloisMahdal:
`qa_verified`: Tested with `0.7.0-0.201711071815Z.PR19`. Looks fine, the downstream test passed.

pirat89 (approved):

PR #20, opened 2017-09-21
Texts editing

jhornice:


RHEL6_7/services/openldap/check
```diff
@@ -55,22 +55,22 @@ If you are using the bdb or hdb openldap back ends, it is necessary to update
 the database.
 
 Back up the data:
-`# slapcat -l backup.ldif`
+# slapcat -l backup.ldif
 
-For more information about slapcat, see the slapcat(8) man pages.
+For more information about slapcat, see the slapcat(8C) man pages.
```
AloisMahdal: The slapcat manpage indeed has `8C` in header. However, it looks like that part is just visual; the real section is still `8`, not `8C`:

    # man 8c slapcat
    No entry for slapcat in section 8c of the manual
    # find /usr/share/man | grep slapcat
    /usr/share/man/man8/slapcat.8.gz
    # man 8 slapcat
    ...(man page is shown)...

(I wonder why the `C` is there in the first place; it looks like a bug to me.)

Also, note that alt. manpage sections are spelled with lowercase letters:

    # man 1P unlink
    No entry for unlink in section 1P of the manual
    # man 1p unlink
    ...(man page is shown)...

They do appear as uppercase in the header but that's because the whole header field is actually uppercase: `UNLINK(1P)`.

RHEL6_7/services/openldap/check
```diff
@@ -55,22 +55,22 @@ If you are using the bdb or hdb openldap back ends, it is necessary to update
 the database.
 
 Back up the data:
-`# slapcat -l backup.ldif`
+# slapcat -l backup.ldif
```
AloisMahdal, line 58: This line and similar lines below actually change behavior: in previous version, the code between the backticks would be actually executed.  It's just due to the fact that all instances of this are prefixed with hash sign, leading to "execution" of comments, ie. not doing anything and just returning empty string.

(Technically, this should have been in different commit.)

RHEL6_7/services/openldap/check
```diff
@@ -132,15 +132,16 @@ fi
...
 Make sure that $ldap_sysconfig ($ldap_sysconfig_new in Red Hat Enterprise Linux 7) is present, and at
 least the SLAPD_URLS variable is set to a valid URL.
-See the slapd(8) man pages, the '-h URLlist' option.
+See the slapd(8C) man pages, the '-h URLlist' option.
 EOF
 
   exit_error
 fi
-log_high_risk "There are crucial changes between Red Hat Enterprise Linux 6 and Red Hat Enterprise Linux 7 openldap configuration. Take the steps recommended in the solution text to ensure the correct functionality."
+<<<<<<< HEAD
```
AloisMahdal: Oops, leftover conflict residue ... aaand we have syntax error :-)

RHEL6_7/services/openldap/check
AloisMahdal: (note that there are more instances of this)

AloisMahdal (changes requested):
Please, make these changes:

 *  Change `8C` manpage reference back to `8`,
 *  remove the `<<<<<<< HEAD` line,
 *  and optionally, move the changes in `RHEL6_7/services/openldap/check` to different commit.

Rest LGTM.

Hint: you can use `find -name check | xargs -L1 file -i | grep shellscript | cut -d: -f1 | xargs -L1 bash -n` to retest syntax, although this does not respect `shopt -s extglob` so it will falsely report line 435 in databases/postgresql; you can ignore that one.

RHEL6_7/services/openldap/check
bocekm: Poor cat, everyone's harsh on her.

RHEL6_7/services/openldap/check
bocekm: Excuse my ignorance, but why would anyone execute a comment?

RHEL6_7/system/RemovableMedia/solution.txt
```diff
@@ -1 +1 @@
-If there are any removable media listed in the /etc/fstab file, it is necessary to add the 'nofail' flag to each one of them. Red Hat Enterprise Linux 7 will fail to boot if the medium is not present and 'nofail' is not specified. It is not possible to verify this automatically with 100% confidence, so check your /etc/fstab file.
+If there are any removable devices listed in the /etc/fstab file, it is necessary to add the 'nofail' flag to each one of them. Red Hat Enterprise Linux 7 will fail to boot if the device is not present and 'nofail' is not specified. It is not possible to verify this automatically with full confidence, so check your /etc/fstab file.
```
bocekm, line 1: You replace `media` -> `devices` here. Shouldn't it be replaced in the module's INI too?

RHEL6_7/system/grubby/module.ini
```diff
@@ -1,6 +1,6 @@
 [preupgrade]
 check_script = check
-content_description = A workaround for crashing grubby. The module fills initrd into the /boot/grub/grub.conf file by the postupgrade script.
-content_title = The grubby tool
+content_description = A workaround to prevent crashing of the grubby tool. The module fills initrd into the /boot/grub/grub.conf file by the postupgrade script.
```
bocekm: Per documentalists guidelines: `tool` -> `utility`

RHEL6_7/system/scl-collection/module.ini
```diff
@@ -1,7 +1,7 @@
 [preupgrade]
 content_title: Red Hat Software Collections
 author: Petr Hracek <phracek@redhat.com>
-content_description: The module checks if RHSCL are installed.
+content_description: The module checks if RHSCL is installed.
```
bocekm: RHSCL is nothing specific. It's more like a named group of packages, and there are many of these _groups_ available for installation. So I would probably add `any` -> `The module checks if any RHSCL is installed.`

RHEL6_7/system/yaboot/solution.txt
```diff
@@ -1,4 +1,4 @@
 The module is valid only on the 64-bit PowerPC architecture and it checks if Yaboot is installed.
 
-Yaboot cannot boot the Red Hat Enterprise Linux 7 system. The Red Hat Upgrade Tool will install GRUB 2 during the upgrade. Create the /etc/default/grub file before running `redhat-upgrade-tool` to ensure the correct GRUB 2 configuration. A misconfiguration might make the system unbootable.
+Yaboot cannot boot the Red Hat Enterprise Linux 7 system. Red Hat Upgrade Tool will install GRUB 2 during the upgrade. Create the /etc/default/grub file before running `redhat-upgrade-tool` to ensure the correct GRUB 2 configuration. A misconfiguration might make the system unbootable.
```
bocekm: 1) I'm simply curious. Why not having _the_ here? We commonly use "The Preupgrade Assistant" for example.
2) Please replace the backticks with single quote marks.

RHEL6_7/backup/UntrackedFiles/solution.txt
```diff
@@ -6,10 +6,10 @@ Some user data, such as user home directories and temporary files, are not track
 
 * The [link:./kickstart/untrackedexpected] file lists the regular files and symlinks created by runtime system operations (for example, handling runlevels, alternatives, and active SELinux modules). Most likely you do not need to care about them, the list is available just for the sake of completeness.
 
-* The [link:./kickstart/untrackedtemporary] file lists all temporary local files on the system that will not be migrated. This is essentially everything in the /cgroup/, /tmp/, and /var/ directories. Most likely you do not need to care about them, the list is available just for the sake of completeness.
+* The [link:./kickstart/untrackedtemporary] file lists all temporary local files on the system that will not be migrated. This is everything in the /cgroup/, /tmp/, and /var/ directories. Most likely you do not need to care about them, the list is available just for the sake of completeness.
 
 It is recommended that you back up all the data before proceeding with the upgrade to Red Hat Enterprise Linux 7. This data might be quite large.
```
bocekm, line 11: Hm, this file seems to mix and match migration vs. upgrade. @pirat89, isn't this module useful just for migration?

RHEL6_7/databases/postgresql/check
```diff
@@ -225,15 +225,15 @@ EOF
     append_to_solution <<EOF
 * Even if the postgresql-server is probably configured correctly, we are unable
   to say for 100% that the server is not used in some specific way on your
-  machine, so check that the
+  computer, so check that the
```
bocekm: Here you replace `machine` -> `computer` however on the other places, you leave the `machine`. I'm for leaving `machine` even here. To me, `computer` evokes something related to hardware. But I suspect that most customers are having virtual machines nowadays. Maybe just `system` would be even better, but that's probably something I would google first to see what's the commonly used term.

RHEL6_7/drivers/ObsoletePOWER6Processors/check
```diff
@@ -12,7 +12,7 @@ ARCH=`arch`
 if [ $ARCH == 'ppc64' ]; then
     grep '^cpu.*POWER6' /proc/cpuinfo
     if [ $? -eq 0 ]; then
-        log_extreme_risk 'POWER6 processors are not supported on the PowerPC and PowerPC64 architectures in Red Hat Enterprise Linux 7.'
+        log_extreme_risk 'POWER6 processors are not supported on the PowerPC and 64-bit PowerPC architectures in Red Hat Enterprise Linux 7.'
```
bocekm: I would be a bit confused here. I guess that by PowerPC you mean 32-bit PowerPC. So isn't it easier to write `on the the 32-bit and 64-bit PowerPC`?

RHEL6_7/services/openldap/check
AloisMahdal: That's my point; the code was never supposed to be executed.  We're just lucky that the code actually happens to be interpreted as comment.

RHEL6_7/storage/cryptsetup-luks/solution.txt
```diff
@@ -4,13 +4,13 @@ created by using the libgcrypt present in Red Hat Enterprise Linux 6 cannot be u
...
+In case the preupgrade script detects the Whirlpool hash in one of the volumes
 involved in the system boot (any volume listed in the /etc/crypttab configuration file),
 it is considered to be a blocking event for the in-place upgrade to continue. We strongly
-recommend to rehash the LUKS header using the cryptsetup-reencrypt package added to
-Red Hat Enterprise Linux 6.6 and later.
+recommend to rehash the LUKS header by using the cryptsetup-reencrypt package added to
+Red Hat Enterprise Linux 6.6 and later versions.
 
 A command for rehashing the LUKS header without changing the master key:
 `# cryptsetup-reencrypt -h sha1 --keep-key <path_to_luks_device>`
```
bocekm: Probably leftover backticks here. Check them out also in `RHEL6_7/system/ca-certificates/solution.txt` and in `RHEL6_7/networking/iptables/solution.txt`.

bocekm (changes requested):

RHEL6_7/services/openldap/check
pirat89: Just want to point out, that actually this is not comment, but text representing the command that the root should do. The idea was use the markdown notation I guess.

RHEL6_7/backup/UntrackedFiles/solution.txt
pirat89: @mbocek: It is informational module used for both modes - upgrade and migration. The path of files in `kickstart` is used just because it is the only usable directory where these files can be stored and kept inside the tarball.  It is related to the https://github.com/upgrades-migrations/preupgrade-assistant/issues/181 RFE for P-A.

bocekm (approved):

AloisMahdal (approved):

AloisMahdal:
qa_ack: just re-run suite for modules. (At least smoke test, which tests syntax errors.)

PR #29, opened 2017-10-24
Add missing backslash

AloisMahdal:


pirat89:
Deja vu. I see that for the 3rd time.

pirat89:
:D commit ae7bc2c69. We have to do reviews more carefully.

pirat89 (approved):

PR #30, opened 2017-10-25
openssh/sysconfig: correct logic and regexps

pirat89:
Previously the module assumed that only problem inside the config file
would be using of the *export* command on lines as: 
    export VARIABLE=VALUE
and in case that files like that don't exist, exit with PASS status.
But the file can contains ANY executable code, so the check was wrong.

Now the PASS or FIXED result is achieved only in case when we can 
ensure the content of the file produce same configuration on both
systems (original and upgraded).

Additionaly, in case we cannot ensure the file is correct, it is copied
into dirtyconf and admin is instructed to verify (&modify) it move it
to cleanconf if the new configuration should be applied automatically
on the terget system.
  

- additionaly it is corrected syntax error in previous solution
- texts have been updated appropriately

### DIFF
--- a/RHEL6_7/services/openssh/sysconfig/check
+++ b/RHEL6_7/services/openssh/sysconfig/check
@@ -4,31 +4,66 @@
 
 #END GENERATED SECTION
 
+back_up_sysconfig() {
+  mkdir -p $(dirname "$2")
+  cp -a "$1" "$2"
+}
 
 SYSCONFIG_FILE="/etc/sysconfig/sshd"
-CLEAN_SYSCONFIG_FILE="$VALUE_TMP_PREUPGRADE/cleanconf/$SYSCONFIG_FILE"
+CLEAN_SYSCONFIG_FILE="$VALUE_TMP_PREUPGRADE/cleanconf$SYSCONFIG_FILE"
+DIRTY_SYSCONFIG_FILE="$VALUE_TMP_PREUPGRADE/dirtyconf$SYSCONFIG_FILE"
 
-mkdir -p $(dirname $CLEAN_SYSCONFIG_FILE)
-cp -a $SYSCONFIG_FILE $CLEAN_SYSCONFIG_FILE
+# put the config file into our dir and possible changes made here. Then it will
+# be moved to expected directory (clean|dirty)conf ...
+SYSCONFIG_TMP="$PWD/sshd.tmp"
+back_up_sysconfig "$SYSCONFIG_FILE" "$SYSCONFIG_TMP"
 
-if grep -q "^[[:space:]]*export[[:space]]" $SYSCONFIG_FILE; then
-    msg="The $SYSCONFIG_FILE file will not be a shell script in Red Hat"
-    msg+=" Enterprise Linux 7 anymore, so all 'export VARIABLE=VALUE' have to"
-    msg+=" be changed to 'VARIABLE=VALUE'."
+# This is checker, whether the $SYSCONFIG_FILE could contains anything, that
+# would cause different output on the upgraded system; Like executable code
+# can't be presented, additional variables, ... e.g.:
+#  var=$(...)
+#  var="String $var2 string"
+#  var=string #comment
+#  ...
+grep -v "^[[:space:]]*#" $SYSCONFIG_FILE | grep -v "^[[:space:]]*$" \
+    | grep -qiEv '^[[:space:]]*(export[[:space:]]*)?[[:alnum:]_]+[[:space:]]*=[^][\$`()!:;#]*$'
+SUSPICIOUS=$?
+
+if grep -q "^[[:space:]]*export[[:space:]]" $SYSCONFIG_FILE || [ $SUSPICIOUS -eq 0 ]; then
+    msg="In Red Hat Enterprise Linux (RHEL) 7, the sshd config file"
+    msg+=" $SYSCONFIG_FILE is no longer a shell script as it was in RHEL 6."
+    msg+=" With the introduction of systemd it has become an environment"
+    msg+=" file for the sshd systemd service, which has a KEY=VALUE syntax."
+    msg+=" Thus, the config file cannot contain any executable code or even"
+    msg+=" lines like 'VARIABLE=VALUE #comment'. Otherwise, unexpected"
+    msg+=" behavior could occur on the target system. The complete file"
+    msg+=" syntax is documented at [link:https://www.freedesktop.org/software/systemd/man/systemd.exec.html#EnvironmentFile=]"
     echo -e "$msg" >> solution.txt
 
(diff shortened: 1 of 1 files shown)


pirat89:
~~Do not merge yet - See PR #14. This module will be discussed yet, if this will be final solution it will be modified yet.~~

jmazanek (dismissed):

pirat89:
The module specification
========================
Root cause
-----------
On RHEL 7 the _sshd_ config file [0] is no longer a shell script as it was on RHEL 6 - with the introduction of systemd it has become an environment file for _sshd_ systemd service which has a KEY=VALUE syntax [2].

short story long
----------------

The module is applied when openssh-server is installed. In that case, back up [0] to:

- [1] when result is fixed or pass (IOW when user do not need pay          
          atteantion to the config file.                                       
- [3] otherwise as action is needed.                                       
                                                                               
When export command is used inside [0] and we can ensure it is only presence 
of shell "code", print solution about changes in RHEL 7 and try to remove    
export command automatically from [1]. When "sed" utility sucesses, log info 
and set result fixed. In case that sed fails, require manual action from     
user (high risk + exit_fail).

In case we cannot ensure that there is not another code or generally,
that produced output between original and new system will be different,
always require manual action from user (false positives are expected
too here). Here is small set of lines, that are evaluated as suspicious
(export command has no effect in this case, still will be removed when
it is possible as described above):

    export VAR=NAME # comment - the comment will be evaluated as string on rh7
    export VAR="Name $variable something"
    VAR=$(...)
    export VAR="This is #1 false positive here"
    export VAR="This is 2nd false positive here;"
    export VAR="This is (3rd) false positive here"
    myfunc() { ... }

...and others. The only goal is to be sure the config file produce same
configuration/output on the upgraded system.


long story short
----------------

    INST | EXP | CODE || BAK   | FIX | RESULT
    ==================||========================
     no  | -   | -    || -     | -   | not_applicable
     yes | no  | no   || clean | -   | pass
     yes | yes | no   || dirty | no  | needs_action (untestable probably)      
     yes | yes | no   || clean | yes | fixed
     yes | no  | yes  || dirty | -   | needs_action
     yes | yes | yes  || dirty | no  | needs_action (untestable probably)      
     yes | yes | yes  || dirty | yes | needs_action


legend:

  - `INST`: openssh-server is installed
  - `EXP`:  The config file [0] uses export command to set and export environment variables for "relative" utilities
  - `BAK`:  back up [0] to [1] (clean) or [3] (dirty)
  - `FIX`:  the "export" command has been removed from [1] (sed returns 0) (correction: export command in case a line has the format "export VAR=VAL")
  - `CODE`: suspicious line has been found (includes possible false positives)


[0] /etc/sysconfig/sshd
[1] /root/preupgrade/cleanconf/etc/sysconfig/sshd
[2] https://www.freedesktop.org/software/systemd/man/systemd.exec.html#EnvironmentFile=
[3] /root/preupgrade/dirtyconf/etc/sysconfig/sshd

AloisMahdal:
@pirat89 I like how you're adopting the case study format.  This really helps to get the gist of how the tests will look.  However, I have small note: In the table it really helps to distinguish between input variables (INST, EXP nad CODE in above) and behaviors (FIX, BAK and RESULT) by keeping first group to the left, the second to the right and delimiting them visually.

Once you're at the point where you posted it, it should be simple operation in Vim using block operations.

(I took the liberty and fixed it.  That and polished formatting a bit.)

pirat89:
The specification has been modified :-)

pirat89:
Added link to the full description of expected format. (inside generated solution text)

bocekm:
Message I'm getting in the report now:

> **Logs**
> preupg.log.INFO: The 'export' commands will be removed from the /etc/sysconfig/sshd file. preupg.risk.HIGH: We cannot ensure the /root/preupgrade/cleanconf//etc/sysconfig/sshd file is correct. Verify manually its correctness.
> 
> **Remediation description:**
> The /etc/sysconfig/sshd file will not be a shell script in Red Hat Enterprise Linux 7 anymore. As a consequence, it should contain only lines in the format 'VARIABLE=VALUE', so all 'export VARIABLE=VALUE' have to be changed to 'VARIABLE=VALUE' and the config file should not contain any executable code or even lines like 'VARIABLE=VALUE #comment'. Otherwise you can get different results on the upgraded system. Lines in the format 'export VARIABLE=VALUE' will be transformed into the expected format automatically. You can find the full format description here: https://www.freedesktop.org/software/systemd/man/systemd.exec.html#EnvironmentFile=

I have two comments on that:
1. In case the result is NEEDS_ACTION we can't ensure the config file correctness, so we should IMO copy it to the `dirtyconf` folder and tell admin to move it to `cleanconf` after they verify it complies with the file syntax.
In case the result is FIXED the config file should go directly to the `cleanconf` folder.
2. I'm missing a clarification in the message on what the Preupgrade Assistant does and mainly, what the admin should: As an admin, I would be confused:
   - "Why was the config file copied to the `cleanconf/dirtyconf` folder?"
   - "Should I fix the copied file or the original? I don't know what will happen during the upgrade - will the original or the copied be used on the upgraded system?"

**My message proposal:**
_On Red Hat Enterprise Linux (RHEL) 7 the sshd config file /etc/sysconfig/sshd is no longer a shell script as it was on RHEL 6 - with the introduction of systemd it has become an environment file for the sshd systemd service which has a KEY=VALUE syntax. Thus, the config file can't contain any executable code or even lines like 'VARIABLE=VALUE #comment'. Otherwise, unexpected behaviour could occur on the upgraded system. The complete file syntax is documented here:
 [link:https://www.freedesktop.org/software/systemd/man/systemd.exec.html#EnvironmentFile=]
The Preupgrade Assistant:_
- _made a copy of the file: /root/preupgrade/**--A--cleanconf--B--dirtyconf**/etc/sysconfig/sshd_
- _removed the 'export' command from each line in the format 'export VARIABLE=VALUE'._

_**--A--Verify that the content of the file copy satisfies the systemd environment file syntax.
--B--Update the file copy to satisfy the systemd environment file syntax and move it then to the /root/preupgrade/cleanconf/etc/sysconfig/ folder.**
The file under the cleanconf directory structure will be used during the upgrade._

See the bold above. **--A--** version is for the `fixed` result and **--B--** is for the `needs_action` result.

AloisMahdal:
`qa_ack`: write test based on specification posted above.

Concentrate on template (or rpm-distribution) sysconfig file: if the most trivial changes are done on default sysconfig file, only then should module attempt to fix them;  otherwise it should require manual migration.

(Consider if this method could be re-used for other files in sysconfig.)

bocekm (approved):

jmazanek (approved):
Tested on my local machine. All cases (e.g config syntax valid for RHEL 7 | valid,  except export keyword | other charachers/invalid) end with the correct result.
Code looks ok

AloisMahdal:
Verified, thanks!

PR #32, opened 2017-10-27
Support for assessment format version

### LINKED ISSUE #31: Support for assessment format version: stage 1 (p-a-m)
This is part of a greater change that needs to be done in all three codebases in 3 stages:

  * [ ] 1. **modules** (this ticket),
  * [ ] 2. preupgrade-assistant ([#335][s2]),
  * [ ] 3. redhat-upgrade-tool ([#31][s3]).

  [s2]: https://github.com/upgrades-migrations/preupgrade-assistant/issues/335
  [s3]: https://github.com/upgrades-migrations/redhat-upgrade-tool/issues/31

Below is the complete description:

---

Assessment, as a data object does not currently have any concept of versioning.  Any tools that further work with it (redhat-upgrade-tool, preupg-kickstart-creator, preupg UI...) simply have to assume that they "understand" the data.

The solution is to introduce assessment format version.   The flow is:

 *  Module set declares its abability to produce assessment in version $V.  This can be done
    by means of key/value pair in properties.ini

        assessment_format_version = 0.1

 *  Assisstant launches modules and passes this information as part of assessment
    somewhere under (/root/preupgrade as well as in tarball).

 *  *redhat-upgrade-tool* hardcodes range of versions (actually just one particular version)
    it's able to "understand" and compares this against assessment version.  If the versions
    match, it can continue.  If the versions don't match or assessment version cannot be
    determined, it advises user to install respective module set and re-run assessment.

pirat89:
We want to make sure that the Red Hat Upgrade Tool and the Preupgrade Assistant (the other tools) will be able to "understand" the data generated by the Preupgrade Assistant Modules (the Modules). For this purpose the Modules will provide a new key inside the _properties.ini_ file: `ast_fmt_version`. It's value will contain version in X.Y format, with semantic:
- X will be bumped when "something" in the Modules changes in such a way that the other tools need to be updated to cope with the change,
- Y will be bumped, when just "something" new has been added to the Modules, but the other tools do not require any change because of this addition in order to work as expected with the data generated by the Modules.

The "something" will be specified later in docs. Currently we can say, that it is everything that is parsed, read, required by the other tools for their correct functionality, but it doesn't include e.g. specific pre-upgrade or post-upgrade scripts or texts in the Preupgrade Assistant report.
  ...
But we can say, that it includes e.g.:
  dirtyconf directory
  hooks directory
  kickstart/RHRHEL7rpmlist_replaced file; that is used when kickstart
                                          is generated
  ...

Resolves #31

### DIFF
--- a/RHEL6_7/properties.ini
+++ b/RHEL6_7/properties.ini
@@ -1,3 +1,4 @@
 [preupgrade-assistant-modules]
 src_major_version=6
 dst_major_version=7
+assessment_format_version=0.1


pirat89:
Note: do not merge this yet

jmazanek (dismissed):
I agree with the suggested concept of version control as such. 
Perhaps the name of the new .ini file directive should be more intuitive. But this should not be a big issue, as long as it is well explained in the documentation.

RHEL6_7/properties.ini
```diff
@@ -1,3 +1,4 @@
 [preupgrade-assistant-modules]
 src_major_version=6
 dst_major_version=7
+ast_fmt_version=0.1
```
bocekm: `ast` is really bad abbreviation IMO. I had a hard time to figure out what the `ast` could mean. Please come up with a name that quickly gives you the idea what it stores.
I vote for `assessment_format_version` or (if you like shorter names; P.S. I don't) `assessment_fmt_ver`.

bocekm (changes requested):

AloisMahdal:
Just note about the versioning strategy; what I've suggested in issue #31 is actually much simpler than OP:

 *  SemVer but only `major.minor` (no `.patch`),
 *  and while version is `0.*`, we don't have to distinguish between "small" and "big"
    updates.

IOW, we will simply bump the `minor` number with *any* change.  (Note that since the only supported workflow is to always re-run assessment right before upgrade, we should never need to support multiple versions.)

Once we are confident we have stable specifications and reasonable coverage tests (that can be wayyy down in future, if we ever reach that point), we can bump to `1.0` and start applying the rules properly.

pirat89:
@bocekm: Done :-)

bocekm (approved):

pirat89:
Closing the PR as after discussion this does not seem relevant. The version of format of assessment will by specified by preupgrade-assistant. Not by modules.

Issue #33, opened 2017-10-31
Informational module: Init scripts to unit files


bocekm:
Provide a guide on transforming init scripts to unit files for systemd.

pirat89:
Resolved by PR #36

PR #36, opened 2017-11-06
initscripts/control: Add msg how to migrate init to unit file

pirat89:
Added short message about replacement of the init by systemd service
manager and link to the guidance how to migrate init scripts to unit
files.

Issue: #33


pirat89:
**Known limitation**: the text will be displayed everytime when the initscripts pkg is installed, but would be nice to print it only in case some custom init script has been detected. As we know that we want to module rewrite completely, I put it just into the solution.txt file, so just text review is needed. Nothing more is expected from this PR.

bocekm (dismissed):

jmazanek (approved):
looks ok

PR #41, opened 2017-11-14
networking/squid

jmazanek:
- Previously in some cases code could
   exit with combination of "$RESULT_INFORMATIONAL"
   and log_*_risk. Cases were also decided in logically
   inconsistent way. With this fix, log and results
   combinations are conforming to the new structure
   Related: PR #218

- Fixing logical flaws in some test statements
  e.g. [ $? -ne 0 ] vs [ $? -eq 0 ]

- Fixing typo causing syntax error


jmazanek:
related preupgrade-assistant PR - https://github.com/upgrades-migrations/preupgrade-assistant/pull/218

pirat89 (approved):
It seems good to me.

PR #47, opened 2017-11-22
Remove links from the risk logs

pirat89:
Logs are processed just as raw strings and are not modified at all. So use of constructions like
  [link:url]
is wrong here and it should not be used.


jmazanek (dismissed):
looks O.K. to me

bocekm (approved):

PR #48, opened 2017-11-22
openssh/sshd: rewrite module + minor changes

pirat89:
See specification in comment below. The module has been refactored at first with minor fixes and the last commit contain fix applied on *refactored* code.

*done:*
 - [x] refactoring
 - [x] fix the logic of the module
 - [x] write specification

### DIFF
--- a/RHEL6_7/services/openssh/sshd/check
+++ b/RHEL6_7/services/openssh/sshd/check
@@ -1,7 +1,6 @@
 #!/bin/bash
 
 . /usr/share/preupgrade/common.sh
-check_applies_to "openssh-server"
 
 #END GENERATED SECTION
 
@@ -9,20 +8,18 @@ check_applies_to "openssh-server"
 CLEANCONF_D="$VALUE_TMP_PREUPGRADE/cleanconf"
 SSHD_CONFIG_FILE="/etc/ssh/sshd_config"
 CLEANCONF_SSHD="${CLEANCONF_D}${SSHD_CONFIG_FILE}"
-
 RESULT=$RESULT_PASS
 
 cp -a --parent $SSHD_CONFIG_FILE $CLEANCONF_D
 
 if grep -q "^[[:space:]]*RequiredAuthentications2" $SSHD_CONFIG_FILE; then
-    solution_file "
-RequiredAuthentications2 is replaced by AuthenticationMethods in $CLEANCONF_SSHD.
-If you want to fix the config file yourself, type:
+    msg="* RequiredAuthentications2 is replaced by AuthenticationMethods in"
+    msg+=" $CLEANCONF_SSHD. If you want to fix the config file yourself, type:"
+    msg+="\n  # sed -i -e 's/^[[:space:]]*RequiredAuthentications2/AuthenticationMethods/i' ${SSHD_CONFIG_FILE}"
+    msg+="\nFor more information about AuthenticationMethods, see the SSHD_CONFIG(5) man page."
+    echo -e "$msg\n" >> solution.txt
 
-# sed -i -e 's/^[[:space:]]*RequiredAuthentications2/AuthenticationMethods/i' $SSHD_CONFIG_FILE
-
-For more information about AuthenticationMethods, see the SSHD_CONFIG(5) man page.
-"
+    #FIXME: fix text and the logic - slight risk + fixed?...
     log_slight_risk "RequiredAuthentication2 will be replaced by AuthenticationMethods."
 
     sed -i -e 's/^\([[:space:]]*\)RequiredAuthentications2/\1AuthenticationMethods/i' $CLEANCONF_SSHD \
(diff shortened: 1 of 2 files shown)


pirat89:
@jmazanek: The issue is related to *risk + exit code* issue. So skip this module now in case you strike this during fixing all risk + exit code issues.

pirat89:
short story long:
-----------------
The module is applied when openssh-server is installed. Unless [0] contains any
of these options:
    - RequiredAuthentications1
    - RequiredAuthentications2
    - AuthorizedKeysCommand
result is PASS as used configuration is compatible with one that can be used
on RHEL 7.

In case the RequiredAuthentications1 is detected, inspection is recommended (slight risk + exit fail) as the option is used for unsupported protocol SSH-1. Result file has to be stored as [2] and with corresponding line commented out.

In case that RequiredAuthentications2 is detected, the option should be replaced by AuthenticationMethods in the modified config [1 or 2]. In case the replacement fails, manual action is required and file is definitely stored as [2] (high risk + exit fail).

When the AuthorizedKeysCommand is detected and when AuthorizedKeysCommandRunAs
    a) is detected, replace the second one by AuthorizedKeysCommandUser
    b) is missing, add line 'AuthorizedKeysCommandUser %u'
in [1 or 2]. When anything fails, the (modified) config file is stored as [2].
    
In case the RequiredAuthentications1 is not detected and no fail is detected the file is stored as [1] with result fixed. When fixing of file fails, it is always stored as [2] and manual action is requested.
    
In case the copying of the file fails, error is logged and then scripts exits with error result.    
    
In case the script doesn't end with error result, postupgrade script
  a) do nothing when [1] exists
  b) back up [0] as [3] and replace it by [2]
In case of (b), do additionaly one more time scan of the [0] and replace/remove
relevant options when they occurs. Just to be sure, the final file doesn't contain any of these.

long story short:
----------------------
```
INST | RA1 | RA2 | AKC | AKCRU | BAK_FAIL | FIX_FAIL || BAK | RESULT    
=====================================================||====================    
 no  | -   | -   | -   | -     | -        | -        || -   | not_applicable   
 yes | -   | -   | -   | -     | yes      | -        || -   | error    
 yes | no  | no  | no  | -     | no       | -        || [1] | pass    
 yes | no  | no  | yes | yes   | no       | no       || [1] | fixed    
 yes | no  | no  | yes | no    | no       | no       || [1] | fixed    
 yes | no  | yes | no  | -     | no       | no       || [1] | fixed    
 yes | no  | yes | yes | yes   | no       | no       || [1] | fixed    
 yes | no  | yes | yes | no    | no       | no       || [1] | fixed    
    
 yes | yes | no  | no  | -     | no       | no       || [2] | needs_inspection 
 yes | yes | no  | yes | yes   | no       | no       || [2] | needs_inspection 
 yes | yes | no  | yes | no    | no       | no       || [2] | needs_inspection 
 yes | yes | yes | no  | -     | no       | no       || [2] | needs_inspection 
 yes | yes | yes | yes | yes   | no       | no       || [2] | needs_inspection 
 yes | yes | yes | yes | no    | no       | no       || [2] | needs_inspection 
    
 yes | no  | no  | yes | yes   | no       | yes      || [2] | needs_action     
 yes | no  | no  | yes | no    | no       | yes      || [2] | needs_action     
 yes | no  | yes | no  | -     | no       | yes      || [2] | needs_action     
 yes | no  | yes | yes | yes   | no       | yes      || [2] | needs_action     
 yes | no  | yes | yes | no    | no       | yes      || [2] | needs_action     
 yes | yes | no  | no  | -     | no       | yes      || [2] | needs_action     
 yes | yes | no  | yes | yes   | no       | yes      || [2] | needs_action     
 yes | yes | no  | yes | no    | no       | yes      || [2] | needs_action     
 yes | yes | yes | no  | -     | no       | yes      || [2] | needs_action     
 yes | yes | yes | yes | yes   | no       | yes      || [2] | needs_action     
 yes | yes | yes | yes | no    | no       | yes      || [2] | needs_action     
    
legend:    
  INST:     openssh-server is installed    
  RA1:      RequiredAuthentications1 detected    
  RA2:      RequiredAuthentications2 detected    
  AKC:      AuthorizedKeysCommand detected    
  AKCRU:    AuthorizedKeysCommandRunAs detected    
  BAK_FAIL: back up failed    
  FIX_FAIL: fix failed    
  BAK:      store (modified) [0] into [1] or [2]
```

[0] /etc/ssh/sshd_config    
[1] /root/preupgrade/cleanconf/etc/ssh/sshd_config    
[2] /root/preupgrade/dirtyconf/etc/ssh/sshd_config    
[3] /etc/ssh/sshd_config.preupg_bp

AloisMahdal:
`qa_ack`: add test based on case table above.

(@pirat89, nice work there!)

jmazanek (dismissed):

AloisMahdal:
I need to clarify few things, @pirat89:

> In case that RequiredAuthentications2 is detected, the option should be replaced by
> AuthenticationMethods in the modified config [1 or 2]. In case the replacement fails,
> manual action is required and file is definitely stored as [2] (high risk + exit fail).

How can I reproduce fail of replacement?  (Also shouldn't that be `error`? AFAIK the replacement reds from /etc and writes in /root/preupgrade.)

> In case the script doesn't end with error result, postupgrade script
> a) do nothing when [1] exists
> b) back up [0] as [3] and replace it by [2]

`[1]` is clean, `[2]` is dirty, so shouldn't it be other way around?

> In case of (b), do additionaly one more time scan of the [0] and replace/remove
> relevant options when they occurs. Just to be sure, the final file doesn't contain any of these.

What should be name of the postupgrade script?

Frankly I have no idea how to reproduce (in a valid workflow) triggering of the second scan.  Also it seems like bad design: do we not trust our own code?


Also, few notes about the case table:

 *  AKCRU does not really change any result, so it's either superfluous or something's missing on
    the behavior side.

 *  Unless this module uses non-standard way (ie. other than `applies_to`), we don't need to have things
    like INST in the table; we won't probably test them anywat (`applies_to` is already covered in preupg
    tests).

AloisMahdal:
Another one:

> In case that RequiredAuthentications2 is detected, the option should be replaced by
> AuthenticationMethods in the modified config [1 or 2]. In case the replacement fails, manual action is
> required and file is definitely stored as [2] (high risk + exit fail).
> [...]
> When the AuthorizedKeysCommand is detected and when AuthorizedKeysCommandRunAs
> a) is detected, replace the second one by AuthorizedKeysCommandUser
> b) is missing, add line 'AuthorizedKeysCommandUser %u'
> in [1 or 2]. When anything fails, the (modified) config file is stored as [2].

so which one is it? 1 or 2?

AloisMahdal:
First test already fails:

    # tail /etc/ssh/sshd_config
    Subsystem	sftp	/usr/libexec/openssh/sftp-server

    # Example of overriding settings on a per-user basis
    #Match User anoncvs
    #	X11Forwarding no
    #	AllowTcpForwarding no
    #	ForceCommand cvs server
    SomeOtherDirective foo_sod
    RequiredAuthentications1 foo_ra1
    RequiredAuthentications2 foo_ra2

and the module throws this error:

    sed: -e expression #1, char 44: comments don't accept any addresses

pirat89:
>and the module throws this error:
>
>`sed: -e expression #1, char 44: comments don't accept any addresses`

I see. My mistake. I removed `s` before regexp to test functionality of the solution when replacement fails and haven't returned it back.

pirat89:
>I need to clarify few things, @pirat89:

>> In case that RequiredAuthentications2 is detected, the option should be replaced by
>> AuthenticationMethods in the modified config [1 or 2]. In case the replacement fails,
>> manual action is required and file is definitely stored as [2] (high risk + exit fail).

>How can I reproduce fail of replacement?  (Also shouldn't that be `error`? AFAIK the replacement reds from /etc and writes in /root/preupgrade.)

You can't, I told you time ago that this is there for completeness to be sure everything is correct. Only way is change of the code for purposes of the test probably or some change of system between the check and action. Yes, you don't want to do that. The reason why we have that there is because we found similar *impossible* in the past, so I did it completele to be sure, that in magic-case nothing wrong happen.

>> In case the script doesn't end with error result, postupgrade script
>> a) do nothing when [1] exists
>> b) back up [0] as [3] and replace it by [2]

>`[1]` is clean, `[2]` is dirty, so shouldn't it be other way around?

Not sure that I understand your point. Usually it should be always in [1], but in case that something fails, the [2] is like seatbelt, because without fixed configuration file on RH7, ssh will not work. It looks redundant, but it is really just to be sure it will be fine.

>> In case of (b), do additionaly one more time scan of the [0] and replace/remove
>> relevant options when they occurs. Just to be sure, the final file doesn't contain any of these.

>What should be name of the postupgrade script?

`openssh.sh` It should be copied by PA automatically, but now I don't see it there, so maybe I will have to add copy from the `check` script

>Frankly I have no idea how to reproduce (in a valid workflow) triggering of the second scan.  Also it seems like bad design: do we not trust our own code?

When somthing failed, you can't trust that the file is correct, right? In that case you have to do scan and try it one more time to fix it. But it's theoretical situation that should not happen ideally.

>Also, few notes about the case table:

> *  AKCRU does not really change any result, so it's either superfluous or something's missing on
    the behavior side.

It will not change result in the meaning of exit status and that's expected - otherwise it would be bug. When AKCRU would't be resolved in the script, you would not be able to connect using ssh. Even the report will be different. However I don't assume we need test content of the message here.

> *  Unless this module uses non-standard way (ie. other than `applies_to`), we don't need to have things like INST in the table; we won't probably test them anywat (`applies_to` is already covered in preupg tests).

As I said, I have it there for completeness. We can sync when you are here in person to discuss it.

jmazanek (dismissed):

AloisMahdal:
OK, testing with *preupgrade-assistant-el6toel7-0.7.2-0.201804031203Z.PR48.el6_8*: 

Most probably a bug: If I don't touch the config at all, it gets copied without changes to dirtyconf.

pirat89:
@AloisMahdal With @tcerna the best QE team!!! Fixed.

jmazanek (dismissed):
Fix looks good to me

jmazanek (approved):

AloisMahdal:
`qa_verified`: Verified with *preupgrade-assistant-el6toel7-0.7.2-0.201804041451Z.PR48.el6_8* via dedicated downstream test.

Note that we have *not* covered cases with FIX_FAIL and BAK_FAIL; these cannot be normally reproduced by user so they are low-pri at best.

PR #50, opened 2018-01-08
Risk_ fix

jmazanek:
This is a first batch of log_*_risk and exit combinations fixes.
Only simple fixes were included. Where there was a need for more complex code change, I did it in separate branch.

### DIFF
--- a/RHEL6_7/networking/iptables/check
+++ b/RHEL6_7/networking/iptables/check
@@ -55,8 +55,8 @@ if service_is_enabled "iptables" || service_is_enabled "ip6tables"; then
     cp $POST_SCRIPT $POSTUPGRADE_DIR/$SCRIPT_NAME
   fi
 
-  log_warning "Read the remediation instructions."
-  exit_informational
+  log_slight_risk "The iptables or ip6tables service is enabled.Read the remediation instructions."
+  exit_fail
 fi
 
 exit_pass
--- a/RHEL6_7/networking/net-snmp/check
+++ b/RHEL6_7/networking/net-snmp/check
@@ -5,14 +5,14 @@
 #END GENERATED SECTION
 
 if service_is_enabled snmpd; then
-    log_slight_risk "Net-SNMP daemon is enabled. See the following knowledge base article for known incompatibilities."
+    log_medium_risk "Net-SNMP daemon is enabled. See the following knowledge base article for known incompatibilities."
     cat <<_EOF_ >solution.txt
 Net-SNMP in Red Hat Enterprise Linux 7 has been updated to version 5.7.2. It includes many fixes and new features.
 In most configurations, no changes to the configuration files are necessary.
 See the following knowledge base article at [link:https://access.redhat.com/site/articles/696163] for any known incompatibilities.
 Retest carefully all applications consuming SNMP data from this system with the updated Net-SNMP package.
 _EOF_
-    exit $RESULT_INFORMATIONAL
+    exit $RESULT_FAIL
 fi
 
 exit $RESULT_PASS
--- a/RHEL6_7/storage/multipath/check
+++ b/RHEL6_7/storage/multipath/check
@@ -13,7 +13,7 @@ mkdir -p $VALUE_TMP_PREUPGRADE/postupgrade.d/multipath
 cp -R postupgrade.d/* $VALUE_TMP_PREUPGRADE/postupgrade.d/multipath
 
 if [ -e "$OLD_MULTIPATH_RULES" ]; then
-	log_warning "Multipath's udev rules file has moved in Red Hat Enterprise Linux 7. The old $OLD_MULTIPATH_RULES file must be removed."
+	log_info "Multipath's udev rules file has moved in Red Hat Enterprise Linux 7. The original $OLD_MULTIPATH_RULES file will be removed by the postupgrade script."
 	FIX=1
 fi	
  
@@ -23,7 +23,7 @@ if [ -f "$MULTIPATH_CONF" ]; then
 		ERR=1
 	fi
 	if grep -q "devices[[:space:]]*{" $MULTIPATH_CONF; then
-		log_high_risk "Multipath matches user device configs with built-in device configs differently in Red Hat Enterprise Linux 7."
+		log_info "Multipath matches user device configs with built-in device configs differently in Red Hat Enterprise Linux 7. This will be taken care of by the postupgrade script."
 		FIX=1
 	fi
 fi
--- a/RHEL6_7/networking/dnsmasq/check
+++ b/RHEL6_7/networking/dnsmasq/check
@@ -146,7 +146,7 @@ def check_dhcp_tags(file_path, buff):
     dhcp_tag_iter = dhcp_tag_pattern.finditer(buff)
 
     for statement in dhcp_tag_iter:
-        log_slight_risk("It looks like you are using DHCP tags ('" + statement.group(0) + "') in \"" + file_path + "\".")
+        log_info("It looks like you are using DHCP tags ('" + statement.group(0) + "') in \"" + file_path + "\".")
         status = EXIT_INFORMATIONAL
 
     if status == EXIT_INFORMATIONAL:
--- a/RHEL6_7/networking/nfsv2/check
+++ b/RHEL6_7/networking/nfsv2/check
@@ -21,7 +21,7 @@ echo "NFS protocol version 2 is not supported in Red Hat Enterprise Linux 7.
 $_nice_separator" > solution.txt
 
 if [[ "$FSTAB" != "" ]]; then
-  log_slight_risk "The /etc/fstab file mounts certain file systems with an unsupported protocol NFSv2."
+  log_info "The /etc/fstab file mounts certain file systems with an unsupported protocol NFSv2."
   echo "The /etc/fstab file mounts these file systems with NFSv2:" >> solution.txt
   echo -e "$FSTAB" | awk '{print "    "$1}' >> solution.txt
   #Change protocol version or remove 'nfsvers' option.
@@ -33,7 +33,7 @@ $_nice_separator" >> solution.txt
 fi
 
 if [[ "$NFSMOUNT" != "" ]]; then
-  log_medium_risk "The /etc/nfsmount.conf file contains settings with NFSv2."
+  log_info "The /etc/nfsmount.conf file contains settings with NFSv2."
   echo "The /etc/nfsmount.conf file contains:" >> solution.txt
   echo -e "$NFSMOUNT
 FIXED: The lines Nfsvers and Defaultvers with value 2 will be commented out.
--- a/RHEL6_7/networking/sendmail/check
+++ b/RHEL6_7/networking/sendmail/check
@@ -51,6 +51,7 @@ processing interval is now specified by SENDMAIL_OPTS in the $CONFIG_FILE file. 
 To setup queue processing interval to, for example, one hour, use SENDMAIL_OPTS=\"-q1h\"."
 
   if ! grep -q "SENDMAIL_OPTS=.*-q[0-9smhdw]\+.*" "$CONFIG_FILE"; then
+    log_medium_risk "The mail queue processing interval is not set."
     solution "Your $CONFIG_FILE file does not contain SENDMAIL_OPTS=-qTIME. If you do \
 not specify the queue processing interval, the mail queue will be processed only once \
 during the sendmail startup, and further emails will not be processed or \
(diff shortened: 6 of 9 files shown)


AloisMahdal:
I find the way you split the change into commits confusing.  It's not useful to split commits just based on *location*.  What should bind changes together is the type and *purpose* of the change, not locality.  This should IMO have been one commit, or if it makes sense and preserves code integrity, commit eg. risk-related change and exit-related change separately.

Also commit message should have the most useful part in first line (similar to Subject line).  Paths to edited file(s) are useless since this is already included in the commit by definition and any tool can already show that information if needed.  (Oh and by the way you also have them wrong at leas in one place. :-))

Ie. commit message like

    packages/ReplacedPackages

     -  blah blah

is not well-formed: it's useless when wieving from list and even wieved in full, it does not tell much.  A better way would be:

    Fix usage of exit/risk/log API

    Since preupgrade-assistant's PR#218[1], the right way is to do blah blah blah. This
    commit updates the API usage in first batch of modules.

      [1]: https://github.com/upgrades-migrations/preupgrade-assistant/pull/218

(Note that the content of the above example is most probably incorrect).

I **highly** recommend [How to Write a Git Commit Message][1] by Chris Beams.

  [1]: https://chris.beams.io/posts/git-commit/

pirat89:
@AloisMahdal just about the start of the title line: In case you are working with specific module, it is helpful to introduce commit with the name of the module, because I immadiately know, which modul has been changed in that way only to see the title. Similar situation is when you are working with modules under small specific group (e.g. mysql). OTOH it doesn't make sense to write something like that when you are providing generic changes between many modules.

I guess that rebase is still expected and this split is here mainly because the work has been done over longer time period, so commits have been generated continuosly with expectation of rebase later. Just about expected rebase, in that case the name of module is used as start of title line, you can do easier squash of commits for specific module.

RHEL6_7/packages/ReplacedPackages/check
```diff
@@ -318,7 +318,7 @@ echo -n "
 
 [ $notprovided -eq 1 ] && [ -z "$other_repositories" ] && statuscode=$RESULT_FIXED
 
-[ $found -eq 1 ] && log_slight_risk "\
+[ $found -eq 1 ] && log_info "\
```
pirat89: Not sure whether this should be just info message always. In case there are packages from outher channels (not-base), these packages willl not be installed automatically usually because the repository is not available in the moment of upgrade.

In addition there can be set `$RESULT_FIXED` but risks are set above already. So this still can be evaluated as error by PA.

RHEL6_7/storage/multipath/check
```diff
@@ -13,7 +13,7 @@ mkdir -p $VALUE_TMP_PREUPGRADE/postupgrade.d/multipath
 cp -R postupgrade.d/* $VALUE_TMP_PREUPGRADE/postupgrade.d/multipath
 
 if [ -e "$OLD_MULTIPATH_RULES" ]; then
-	log_warning "Multipath's udev rules file has moved in Red Hat Enterprise Linux 7. The old $OLD_MULTIPATH_RULES file must be removed."
+	log_warning "Multipath's udev rules file has moved in Red Hat Enterprise Linux 7. The original $OLD_MULTIPATH_RULES file will be removed by the postupgrade script."
```
pirat89: We can't use `log_warning` in this meaning. Purpose of this function is more of development character. It should be replaced by something else. Same for use of `log_warning` below.

RHEL6_7/storage/multipath/check
```diff
@@ -23,11 +23,11 @@ if [ -f "$MULTIPATH_CONF" ]; then
 		ERR=1
 	fi
 	if grep -q "devices[[:space:]]*{" $MULTIPATH_CONF; then
-		log_high_risk "Multipath matches user device configs with built-in device configs differently in Red Hat Enterprise Linux 7."
+		log_warning "Multipath matches user device configs with built-in device configs differently in Red Hat Enterprise Linux 7. This will be taken care of by the postupgrade script."
```
pirat89: You see here used the `log_high_risk` originaly with `$RESULT_FIXED` later.  It smells like the unreliable solution and in that case the result fixed should not be used, whether the solution is not reliable. Same for `log_warning` - doesn't make sense to have log_warning combined with `$RESULT_FIXED`

RHEL6_7/storage/multipath/check
```diff
@@ -23,11 +23,11 @@ if [ -f "$MULTIPATH_CONF" ]; then
...
 		FIX=1
 	fi
 fi
-log_medium_risk "The number of default configuation settings have changed in Red Hat Enterprise Linux 7. Verify that your devices are still optimally configured."
+log_warning "The number of default configuation settings have changed in Red Hat Enterprise Linux 7. Verify that your devices are still optimally configured."
```
pirat89: Doesn't make sense to use `log_warning` when message obviously say that user should do something later. This is obviously `log_medium_risk`.

RHEL6_7/databases/postgresql/check
```diff
@@ -236,7 +236,7 @@ EOF
   See the upstream article for pg_dumpall:
   [link:$UPSTREAM_DOC_PGDUMP]
 EOF
-    log_slight_risk "We cannot tell for 100% that the system will be in-place upgradeable."
+    log_info "We cannot tell for 100% that the system will be in-place upgradeable."
     DIFFERENT_USAGE_OK=yes
     return $RESULT_PASS
```
pirat89: It seems that `$RESULT_PASS` smells here. Read the two messages. I am thinking even whether we can ends with `$RESULT_PASS` ever, because it seems more like that this test is about `$RESULT_FAIL` and `$RESULT_INFORMATIONAL` as in case of pass, you will not see the solution text even. But in that case it is bug that it is already part of upstream for a long time.

RHEL6_7/databases/postgresql/check
```diff
@@ -340,7 +340,7 @@ EOF
     # Even if user misconfigured the pg_hba.conf contents, most likely he did
     # not changed the 'postgres' role.  Thus we can assume he is Superuser (with
     # slight risk).
-    return $RESULT_PASS
+    return $RESULT_FAIL
```
pirat89: according to the message it seems that there should be set even high risk. As I read it properly, this module seems to need even separate pull-request for all changes.

RHEL6_7/networking/net-snmp/check
```diff
@@ -12,7 +12,7 @@ In most configurations, no changes to the configuration files are necessary.
 See the following knowledge base article at [link:https://access.redhat.com/site/articles/696163] for any known incompatibilities.
 Retest carefully all applications consuming SNMP data from this system with the updated Net-SNMP package.
 _EOF_
-    exit $RESULT_INFORMATIONAL
+    exit $RESULT_FAIL
 fi
```
pirat89, line 17: It's ok. But what do you think about use of `log_slight_risk` above? I am not familiar with incompatibilities, so not sure whether it should be slight or medium risk. You can keep it as it is, but maybe we should create somewhere note about that to investigate it later.

RHEL6_7/networking/iptables/check
```diff
@@ -55,8 +55,8 @@ if service_is_enabled "iptables" || service_is_enabled "ip6tables"; then
     cp $POST_SCRIPT $POSTUPGRADE_DIR/$SCRIPT_NAME
   fi
 
-  log_warning "Read the remediation instructions."
-  exit_informational
+  log_slight_risk "Read the remediation instructions."
```
pirat89: The message is too generic. It should be changed to something like:
- *The iptables or ip6tables service is enabled. Read the remediation instructions.*

What do you think?

pirat89 (dismissed):

RHEL6_7/databases/postgresql/check
jmazanek: I think this is a question of a problem scope. Do we define `$RESULT_PASS` as a total success of the upgrade/migration process, without any chance of fail? Or do we define it as a success in the scope, what we are able to provide with preupgrade assistant? If the latter is a case, then I think the handling of the actual user data content of the database is outside of this scope and `$RESULT_PASS` is justified.

RHEL6_7/databases/postgresql/check
jmazanek: From the message it seems, that the severity of the issue depends on the 'postgres' role. With the default configuration, severity is low. I agree with module rewrite in separate pull request where, the logic to determine the 'postgres' role will be added.

RHEL6_7/networking/iptables/check
jmazanek: Yes, I will add that info

RHEL6_7/networking/net-snmp/check
jmazanek: Based on the KB and snmpd.conf man page, I assume it can cause the issue in certain corner cases. It can potentially affect some associated sysadmin scripts. I will change the risk to `log_medium_risk`

RHEL6_7/packages/ReplacedPackages/check
jmazanek: I agree. The not base repositories should be handled with a separate logic cycle. I will fix it in separate PR

RHEL6_7/storage/multipath/check
jmazanek: OK. I wasn't aware of this meaning of log warning. I will change it.

RHEL6_7/storage/multipath/check
jmazanek: Same as the above

RHEL6_7/storage/multipath/check
jmazanek: Same as the above

RHEL6_7/databases/postgresql/check
pirat89: Agree, it doesn't matter in this PR. But we should look at it in future.

AloisMahdal:
@pirat89,

> it is helpful to introduce commit with the name of the module, because I immadiately
> know, which modul has been changed in that way only to see the title

You already know that from the data in the commit.  Repeating it in the title is useless at best; in fact it makes the *other* part of the title harder to read (eg. for people who are able to get that info from git itself).  Also the info is now subject to human factor; just look at the commits in this PR; the info is actually *incorrect* (yet it's counter-productive to expect commiter to focus on getting *this* right).  Lastly, the whole thing falls apart anyway once you make commit in 2 or more modules.

Your *pro* argument only applies when the auditor is lazy *and* does not know their tools; and even in that case it's pretty questionable.

About the second part; I can't see how expectation of rebase, or the way how commits were created is relevant to the review or to my feedback.

AloisMahdal:
> Lastly, the whole thing falls apart anyway once you make commit in 2 or more modules.

Oh and of course it also falls apart if you reorganize or rename the files.

pirat89:
@amahdal:
> About the second part; I can't see how expectation of rebase, or the way how commits were created is relevant to the review or to my feedback.

It was explanation, that this PR is still more like WIP then something what is prepared for review.

> Oh and of course it also falls apart if you reorganize or rename the files.

This is really good point.

bocekm:
@AloisMahdal, thanks for the [How to Write a Git Commit Message](https://chris.beams.io/posts/git-commit/) by Chris Beams. Worth reading and sharing.

jmazanek:
included those of the @pirat89 comments that do not require the separate PR into the code

bocekm:
@jmazanek what will you do with the comments that require a separate PR? To keep track of that, will you create a GitHub issue, or JIRA task? Either is ok with me.

pirat89:
@jmazanek please, create issue for that with checkbox-list of modules that should be fixed separately over time. We should keep it public.

jmazanek:
@bocekm  Jira task already exists in backlog RHELLEAPP-594

jmazanek:
@pirat89 @bocekm I suggest merging this PR as it is. The comments regarding postgresql and ReplacedPackages, while valid, are in their nature an extension ( additional code logic requested - but we don't know, when we will have a time and resources to do them), but changes already done in this PR are not harmful. Reverting them would definitely cause the error ( with a new return code schema as defined in https://github.com/upgrades-migrations/preupgrade-assistant/pull/218 )

pirat89:
@jmazanek: Changes *MAY* not be harmful but in the nature they didn't resolve anything. Anyway, I would not say that sentence like: **We cannot tell for 100% that the system will be in-place upgradeable.** is really just informational. That's more from my point like slight risk when I am reading that, but it's arguable. The checks and results in postgresql module needs to be resolved better IMO as they don't look to be consistent.

The commit messages of both commits (for postgresql and ReplacedPackages) says: `fixing log & exit status combination` but actually it's not true as we now it doesn't fix anything and contain just change as both would still ends with error code with the new schema. When we want to fix all modules, we will go through the modules that have not been fixed yet and these commits will do just mess.

jmazanek:
changes to above specified files reverted. @pirat89 you can approve for merge

pirat89 (dismissed):

pirat89 (approved):
lgtm

PR #51, opened 2018-01-10
system/python: handle ownership of directories by more rpms

pirat89:
Previously in case a directory was owned by multiple rpms, script printed warning message with mismatched names of the pkgs. Now it is handled correctly and in case that the directory is owned at least by one native rpm.

refactoring: Replaced backticks by "$(...)"

Example of unwanted wrning message for two packages:
```
preupg.log.WARNING: Package python-backportspython-backports-ssl_match_hostname is not installed on Red Hat Enterprise Linux system.
```

### DIFF
--- a/RHEL6_7/system/python/check
+++ b/RHEL6_7/system/python/check
@@ -10,17 +10,30 @@ if [ -f "$PYTHON_DIRS" ]; then
     rm -f $PYTHON_DIRS
 fi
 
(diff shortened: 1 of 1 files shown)


jmazanek:
LGTM. I've tested the new function and I'm not getting any garbled file names any more.

jmazanek (approved):

PR #52, opened 2018-01-10
Provide directory for third party modules.

pirat89:
We expect that third party modules will be located in the specific directory. In this case the chosen dirname is "3rdparty".

The SPEC file has been modified appropriately.

[0] https://github.com/upgrades-migrations/preupgrade-assistant/pull/348

### DIFF
--- a/RHEL6_7/3rdparty/group.ini
+++ b/RHEL6_7/3rdparty/group.ini
@@ -1,0 +1,2 @@
+[preupgrade]
+group_title = Third party modules
--- a/packaging/preupgrade-assistant-el6toel7.spec
+++ b/packaging/preupgrade-assistant-el6toel7.spec
@@ -85,7 +85,6 @@ install -d -m 755 $RPM_BUILD_ROOT%{_datadir}/preupgrade
 mv LICENSE $RPM_BUILD_ROOT%{_datadir}/doc/preupgrade-assistant-el6toel7/
 mv RHEL6_7-results $RPM_BUILD_ROOT%{_datadir}/preupgrade/RHEL6_7
 
-mkdir -p $RPM_BUILD_ROOT%{_datadir}/preupgrade/RHEL6_7/3rdparty
 rm -rf $RPM_BUILD_ROOT/%{_datadir}/preupgrade/common
 
 # General cleanup


pirat89:
@AloisMahdal Do you think that it's worth to have test for this? whether the directory with the group file exists for the third party packages?

jmazanek (approved):
Looks good to me

AloisMahdal:
> Do you think that it's worth to have test for this?

Adding this to static tests would be pretty trivial, I guess, so why not?...

AloisMahdal:
`qa_ack`: into Sanity/static downstream test, add check that the group.ini file exists.

AloisMahdal:
Only group.xml was added; not group.ini.

    [root@testmachine ~]# rpm -q preupgrade-assistant-el6toel7
    preupgrade-assistant-el6toel7-0.7.1-0.201801101909Z.PR52.el6_8.noarch
    [root@testmachine ~]# rpm -ql preupgrade-assistant-el6toel7 | grep 3rd
    /usr/share/preupgrade/RHEL6_7/3rdparty
    /usr/share/preupgrade/RHEL6_7/3rdparty/group.xml
    [root@testmachine ~]# 

Is this expected? (IIUC group.ini is needed but I might be wrong).

AloisMahdal:
(OTOH, this PR is not `on_qe`, but that might be by mistake.)

pirat89:
> Only group.xml was added; not group.ini.

@AloisMahdal `group.ini` is expected only in the git repository. During build of the package replacement by `group.xml` is expected in the `3rdparty` directory, so what you write is correct. However, in future doesn't matter whether there will be group.ini or xml file (in case that we will remove "compile" of xccdf compose during build phase of rpm package)

AloisMahdal:
It seems I misiunderstood how this is supposed to behave. My understanding was: During runtime, we are re-composing the whole mset once more so that modules added by user inside `3rdparty` are included in the final soup.  AFAIK, in order for this to work, there has to be `group.ini`, or *xccdf-compose* will not add them (I don't recall whether it fails or ignores them).

From that I concluded that the user should see `group.ini` -- which is why we are adding it (in fact `group.xml` would not matter since it gets re-compiled along with all xmls anyway.).

So... which part I got wrong?

AloisMahdal:
Or maybe I just ran the test too early? I just realized `on_qe` is missing, but that could be just that we're still getting used to the workflow.

AloisMahdal:
`qa_verified`: This is now covered in downstream test suite and relevant test has passed with *preupgrade-assistant-el6toel7-0.7.1-0.201801101909Z.PR52.el6_8.noarch*

pirat89:
Thanks Lojzo, merged in el6toel7 branch.

PR #54, opened 2018-01-12
initscripts/control: update preset file

### LINKED ISSUE #53: initscripts/control: update the bundled preset file
The preset file we bundle [0] maybe will be changed. We should keep an eye on that and update it for upgrades to RHEL 7.5 system in that case.

[0] RHEL6_7/system/initscripts/control/90-default.preset

pirat89:
Update bundled preset file to newer one we expect will be used in RHEL 7.5. In addition add there hypervfcopy service as enabled even when it is not part of the preset file on target system. The service is enabled automatically when corresponding device is detected. This change suppress unwanted message in report.

Resolves #53 .


jmazanek (approved):
LGTM

PR #64, opened 2018-02-11
Merge specification of some modules

pirat89:
We have already created specifications of some modules. Some of them accepted favourite *case studies* format, some of them are written with exhausting description only, but still they shoul provide full documentation for purposes of testing.

Part of this PR is modification of the SPEC file, that removes *module_spec* files from RPM, so these files will not be installed on system.

### DIFF
--- a/RHEL6_7/databases/mysql/configuration_changes/module_spec
+++ b/RHEL6_7/databases/mysql/configuration_changes/module_spec
@@ -1,0 +1,36 @@
+- applicable when mysql or (or and? not sure here about behaviour) mysql-libs
+  are installed - at least when any of them is installed, should be result
+  NOT_APPLICABLE (maybe if missing one of them, ask phracek about it)
+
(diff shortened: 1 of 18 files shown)


pirat89:
@AloisMahdal : Lojzo, this is more for you, so please do quick overview, whether some files you want to have modified before merge or we can merge it already. Maybe we should append test, that in RPM are not any *module_spec* files, what do you think?

AloisMahdal:
LGTM; I haven't read them but I guess it's better to have them in and improve them as we write tests.

About the RPM: I don't see any issue with specs being shipped in RPM. (If they really should not be there, it would be trivial to cover this in downstream static test.)

AloisMahdal (dismissed):

pirat89:
Thanks :-) So I will squash bunch of commits and then rebase it to master.

jmazanek (approved):

PR #67, opened 2018-02-22
 system/perl - local dirs only

jmazanek:
- limiting lookup of the perl
   modules to local directories
   only to shorten the execution
   time

### DIFF
--- a/RHEL6_7/system/perl/check
+++ b/RHEL6_7/system/perl/check
@@ -23,6 +23,15 @@ EOM
 
 PERL_DIRS=$(perl -MConfig -e '$,=q{ }; print @Config{installarchlib,installprivlib,installvendorarch,installvendorlib}') \
     || exit_error
+#Limiting the lookup to local directories only
+for dir in ${PERL_DIRS[@]};
+do 
+    fs_type=$(stat -f -L -c %T "$dir")
+    if  echo "$fs_type" | egrep -q 'nfs|nfs4|autofs|cifs|sshfs|fuseblk'; then
+        PERL_DIRS=( "${PERL_DIRS[@]/$dir}")
+        log_info "Directory $dir is not on the local filesystem and it will be skipped in the analysis"
+    fi
+done
 for file in $(find -P $PERL_DIRS -type f -name '*.pm'); do
     RPM=$(rpm -qf "$file")
     if [ $? -ne 0 ]; then
--- a/RHEL6_7/system/perl/check
+++ b/RHEL6_7/system/perl/check
@@ -6,9 +6,28 @@
 #END GENERATED SECTION
 
 FOREIGN_FOUND=0
-function log_file() {
-        printf '%s\n' "$@" >> "$SOLUTION_FILE" || exit_error
-        FOREIGN_FOUND=1
+log_file() {
+    printf '%s\n' "$@" >> "$SOLUTION_FILE" || exit_error
+    FOREIGN_FOUND=1
+}
+is_module_rh() {
+    
+    if [ $# -ne 1 ]; then
+         return 1
+    fi
+    
+    RPM_NAME=$(rpm -qf --qf '%{NAME}\n' "$1")
+    
+    if [ $? -eq 1 ]; then
+        log_slight_risk "The $1 Perl module is not handled by any package."
+        log_file "$1"
+    else
+        is_dist_native $RPM_NAME
+        if [ $? -ne 0 ]; then
+            log_slight_risk "The $1 Perl module was not installed by any package signed by Red Hat."
+            log_file "$1"
+        fi
+    fi
 }
 
 cat > "$SOLUTION_FILE" <<'EOM'
@@ -33,18 +52,7 @@ do
     fi
 done
 for file in $(find -P $PERL_DIRS -type f -name '*.pm'); do
-    RPM=$(rpm -qf "$file")
-    if [ $? -ne 0 ]; then
-        log_slight_risk "The $file perl module is not handled by any package."
-        log_file "$file"
-        continue
-    fi
-    RPM_NAME=$(rpm -q --qf '%{NAME}' "$RPM")
-    is_dist_native "$RPM_NAME"
-    if [ $? -ne 0 ]; then
-        log_slight_risk "The $file perl module was not installed by any package signed by Red Hat."
-        log_file "$file"
-    fi
+    is_module_rh "$file"
 done
 
 test "$FOREIGN_FOUND" = 0 && exit_informational || exit_fail


jmazanek:
This fix should limit the directories for a lookup of perl modules to local file system. There is a native perl way how to stat the filesystems, that would be preferable, however I was not able to interpret it's numerical output.
If someone knows more about the numerical codes of filesystems that perl uses, please let me know and I will change it. Also the list of remote filesystems I used is probably not exhaustive.

jmazanek:
BTW. The Config module contains all the information that was available to the Configure program at Perl build time. Therefore the directories listed in "@Config" array are only a subset of "@INC" array. If it contains remote directories, it means they were added by configure script during package building. So the issue might be somewhere else

RHEL6_7/system/perl/check
```diff
@@ -23,6 +23,14 @@ EOM
 
 PERL_DIRS=$(perl -MConfig -e '$,=q{ }; print @Config{installarchlib,installprivlib,installvendorarch,installvendorlib}') \
     || exit_error
+#Limiting the lookup to local directories only
+for dir in ${PERL_DIRS[@]};
+do 
+    fs_type=$(stat -f -L -c %T "$dir")
+    if  echo "$fs_type" | egrep -q 'nfs|nfs4|autofs|cifs|sshfs'; then
```
bocekm: I can't find a source of all the possible values. @jmazanek , do you have such list?
For example my USB-connected disk is of type `fuseblk` which I even didn't know it exist as a filesystem type.

pirat89 (dismissed):
~This seems completely wrong. Not sure whether there is different behaviour in bash on RHEL 6, but on my Fedora system it will just append suffix "/$dir" to the last item of the array, when $dir is not stored locally on the machine. I don't think that there will be different behaviour on RHEL 6 system, but in case it is, it is for me still wrong solution and should be used something more generic.~ (EDIT: my mistake, it works)

Anyway, when directory should be skipped, it should be reported every time, that it is skipped specific directory for check (use `log_info`). Not sure whether we should added to the current milestone. I prefer to move it to the next one.

RHEL6_7/system/perl/check
jmazanek: No, these are just the most common remote filesystems I could recall.  Didn't consider USB disks, but I can add more

jmazanek:
@pirat89 you read the syntax incorrectly. Construct `${ARRAY[@]/$var}`  is used to remove `$var` from `${ARRAY[@]}`

jmazanek:
@pirat89 @bocekm I will edit the user messages and try to add more FS types ( relevant suggestions are welcomed). We can postpone the merging after the current milestone

RHEL6_7/system/perl/check
pirat89: @jmazanek: Generic solution can be done in similar way how we create list of local executables. Create list of local directories, ignoring symlinks and than check whether directory is part of the list. But it will consume some amount of time, so not sure about benefit at all.

bocekm: Yeah, the Perl module is already the second most time consuming module, so I wouldn't be adding such time consuming stuff.
BTW, I think the module is slow because of how it's written. IOW, it can be greatly optimized IMO. But that's not the subject of this PR.

jmazanek:
Implemented the user messages as per @pirat89 comment. Added `fuseblk` fs type as per @bocekm comment. If we decide to implement more generic solution in the future or other improvements, I will edit it again. But it is not urgent.

jmazanek:
In this enhancement, I did re-factoring of the package check code, in order to reduce an execution time of the module, while keeping the core functionality untouched.
After doing the research I have not found any way how to make the remote directory check more general. On Linux system there is no simple native way how to differentiate between the "local" and "remote" file systems as these are more a usage/custom defined categories. Any other solution would still include parsing arbitrary regexes and couldn't be considered complete. Therefore there is no need to modify this part of the code.

RHEL6_7/system/perl/check
```diff
@@ -6,9 +6,26 @@
...
+if [ $# -ne 1 ]; then
+     return 1
+fi
+RPM_NAME=$(rpm -qf --qf '%{NAME}\n' "$1")
+echo "$RPM_NAME" | grep "is not" >> /dev/null
+if [ $? -eq 0 ]; then
+    log_slight_risk "The $1 Perl module is not handled by any package."
+    log_file "$1"
+else
+    grep "^$RPM_NAME[[:space:]]"  $VALUE_RPM_RHSIGNED > /dev/null
```
pirat89: you should not read this file directly. For that purpose you have function `is_dist_native` which you should use.

pirat89 (dismissed):

RHEL6_7/system/perl/check
jmazanek: There was originally used that function. I have changed it FROM that function because it didn't serve well my purpose :-)  If you look at the `is_module_rh` it is partially based on `is_dist_native`, but tweaked a little bit. Overall it is just read operation ( it doesn't modify anything) and in this way the execution time of the module is cut shorter by 50 % which was a purpose of my fix.

pirat89: as was discussed, it will be used mentioned API function.

RHEL6_7/system/perl/check
```diff
@@ -6,9 +6,26 @@
...
+log_file() {
+    printf '%s\n' "$@" >> "$SOLUTION_FILE" || exit_error
+    FOREIGN_FOUND=1
+}
+is_module_rh() {
+if [ $# -ne 1 ]; then
+     return 1
+fi
+RPM_NAME=$(rpm -qf --qf '%{NAME}\n' "$1")
+echo "$RPM_NAME" | grep "is not" >> /dev/null
```
pirat89: this is redudant line. The grep is not needed to check whether rpm is installed or not because the rpm utility already returns non-zero value when RPM is not installed. remove the line

pirat89 (dismissed):

jmazanek:
@pirat89 your review comments were applied

jmazanek:
Summary of changes & how to test.

Core functionality: 

 This module searches the paths that are available to the Configure program during the build time of perl and verifies if the perl modules contained there are :

a) provided by rpm package

b) signed by Red Hat gpg key

Informative message is printed for the user along with an appropriate risk ( log_slight_risk) if any of the above is false.

What does this PR change:

1) Previously, if one of the paths was a remote directory , the module searched those. This fix exclude the  remote directories accessed over the most common protocols like NFS, CIFS and SSHFS from the search.  Message is printed, informing user, that the given directory has been skipped.

2) While verifying the perl modules, previously this module run redundant rpm queries, which took lot of time.  This fix optimize the queries for the reduced execution time.

How to test it:

a) Omitting remote directory from the analysis

1) Get the list of the perl paths using the following command:

 `perl -MConfig -e '$,=q{ }; print @Config{installarchlib,installprivlib,installvendorarch,installvendorlib}'` 

2) Use one of the directories as a mount point for the remote file system using one of the common protocols (NFS, CIFS or SSHFS)

3) Run the module

4) The directory will be omitted from the analysis and appropriate message will be printed

b) Optimizing execution time

1) Run this build of module

2) Run the original build ( current master)

3) Compare the execution time

4) There should be significant improvement in execution time.

RHEL6_7/system/perl/check
```diff
@@ -6,9 +6,28 @@
...
+    log_slight_risk "The $1 Perl module is not handled by any package."
+    log_file "$1"
+else
+    is_dist_native $RPM_NAME
+    if [ $? -ne 0 ]; then
+        log_slight_risk "The $1 Perl module was not installed by any package signed by Red Hat."
+        log_file "$1"
+    fi
+fi
 }
```
pirat89: Please, keep indentation here, it is hard to read such code.

pirat89 (dismissed):
Just fix the indentation and everything else is ok.

pirat89 (dismissed):
lgtm

AloisMahdal:
`qa_ack`: test manually as described in [comment above][cmt].

  [cmt]: https://github.com/upgrades-migrations/preupgrade-assistant-modules/pull/67#issuecomment-383123165

pirat89 (approved):

AloisMahdal:
Clearing the `qa_ack` from last milestone.

AloisMahdal:
`qa_ack`: test manually as described in [comment above][cmt].

  [cmt]: https://github.com/upgrades-migrations/preupgrade-assistant-modules/pull/67#issuecomment-383123165

drehak:
--test--

drehak:
Execution time greatly reduced as expected.
Tested with:
preupgrade-assistant-el6toel7-0.7.5-0.201808271633Z.PR67.el6_8.noarch
preupgrade-assistant-el6toel7-0.7.5-1.201808221353Z.el6toel7.el6_8.noarch

PR #68, opened 2018-02-27
Rename multiple network devices

jmazanek:
This pull request addresses the problem of the network not coming up after the in-place upgrade of a RHEL 6 machine **with multiple network devices** to RHEL 7.

The module _RHEL6_7/system/initscripts/ifcfg/_ (_xccdf_preupg_rule_system_initscripts_ifcfg_check_) has been updated so that when multiple ethX network devices are present, it generates alternative ifcfg files and udev rules to rename the network devices.

Renaming the network interfaces is automated by a preupgrade script generated by the Preupgrade Assistant module (`/root/preupgrade/preupgrade-scripts/rename_network.sh`) - this script is executed automatically by the redhat-upgrade-tool during the upgrade process, before the user is instructed by the tool to reboot. The user is still expected to review the report and the generated files.

Notes:
- the interfaces are renamed from "eth[0-9]+" to "net[0-9]+", if there are no pre-existing interfaces with "net"  prefix 
- pre-existing interfaces with "net" prefix are taken into account and conflicts are prevented by raising index number of the newly renamed interfaces
- network can be normally restarted with ifup/ifdown scripts after the upgrade
- renaming works correctly with bonded ethX interfaces - the information about which bond device the ethX network interface participates in is configured in the ifcfg-ethX and it is transferred without change to the ifcfg-netX
- occurence of MACADDR directive in ifcfg-ethX would break the network upgrade process, so it inhibits the upgrade

### DIFF
--- a/RHEL6_7/system/initscripts/ifcfg/check
+++ b/RHEL6_7/system/initscripts/ifcfg/check
@@ -35,7 +35,7 @@ else:
 def ifcfg_error():
     """https://bugzilla.redhat.com/show_bug.cgi?id=1111174
 1) kernel names (eth0, eth1, ...)
-a) DEVICE + HWADDR specified in ifcfg files = not safe, udev in rhel7 is not able to swap names (onlky exception here are machines with only one network interface, they should be fine) <- this is only thing which is covered by content which I wrote
+a) DEVICE + HWADDR specified in ifcfg files = not safe, udev in rhel7 is not able to swap names (only exception here are machines with only one network interface, they should be fine)
 b) only DEVICE specified in ifcfg files = not safe, udev will probably name the interface differently in rhel7
 c) only HWADRR specified in ifcfg files = uh, device will be set up correctly according to ifcfg file but name could be different
 2) biosdevname names (em1, p3p4, p3p4_1, ...)
@@ -43,8 +43,7 @@ On dell machines you should be fine in all cases, for others biosdevname was not
 3) custom names (my_little_network_card, ...)
 both HWADDR and DEVICE are specified in this case or you have written your own udev rules.
 This is safe in the case that you use truly unique names, they should not match any other naming methods (kernel, biosdevname or udev in rhel7).
-1a is covered ()
-We should refuse to update also in the 1b case.
+1a and 1b are covered ()
 If you are using 1c case you are used to some level of pain, so it would be enough just to write a warning.
 2 is fine.
 I case 3, to be completely correct you could check that user is not using rhel7 udev naming scheme [1] in rhel6, but I would bet that nobody uses that.
@@ -102,8 +101,18 @@ I case 3, to be completely correct you could check that user is not using rhel7 
                 return True
         return False
 
+    def fix_ifnames():
+        script_result = subprocess.Popen("./preupg_network.sh", shell=True)
+        script_result.communicate()
+        exit_code = script_result.returncode
+        if exit_code == 1:
+            log_extreme_risk("The network interface names on the source system are in an inconsistent state, so the upgrade is not possible")
+        elif exit_code == 2:
+            log_extreme_risk("Comment out the MACADDR directive in the ifcfg files and run the Preupgrade Assistant again to enable the upgrade.")
 
     ethx_with_addr_count = 0
+    ethx_without_addr_count = 0
+    ifnames_to_fix = 0
     warning = False
     for script in ls_scripts():
         full_path = ifcfg_path + script
@@ -124,18 +133,23 @@ I case 3, to be completely correct you could check that user is not using rhel7 
                 if addr:
                     ethx_with_addr_count +=1
                 else:
+                    ethx_without_addr_count +=1
                     log_slight_risk(full_path + " is old style ethX name without HWADDR, its name can change after the upgrade.")
                     warning = True
             elif is_udev(name):
                 log_slight_risk(full_path + " variable DEVICE is similar to udev predictable network naming scheme. It might cause conflicts.")
                 warning = True
 
-        if ethx_with_addr_count > 1:
-            log_medium_risk("You use multiple network devices with old style 'ethX' names.")
+        if ethx_with_addr_count > 1 or ethx_without_addr_count > 1:
+            ifnames_to_fix +=1
             warning = True
-        if ethx_with_addr_count == 1:
-            log_slight_risk("You use one network device with an old style 'ethX' name. This will work as long as you only have one, but it will break with multiple such devices.")
+        elif ethx_with_addr_count == 1:
+            log_slight_risk("You use one network device with an old style 'ethX' name.")
             warning = True
+
+    if ifnames_to_fix > 0:
+        fix_ifnames()
+        log_high_risk("You use multiple network devices with old style 'ethX' names.")
     return warning
 
 if __name__ == "__main__":
--- a/RHEL6_7/system/initscripts/ifcfg/check
+++ b/RHEL6_7/system/initscripts/ifcfg/check
@@ -7,25 +7,25 @@ import re
 
 devel_time = False
 if devel_time:
-	def log_debug(x):
-		print x
+    def log_debug(x):
+        print x
 
-	def log_warning(x):
-		print x
+    def log_warning(x):
+        print x
 
-	def log_error(x):
-		print x
+    def log_error(x):
+        print x
 
-	def exit_pass():
-		sys.exit(0)
-	
-	def exit_fail():
-		sys.exit(1)
+    def exit_pass():
+        sys.exit(0)
 
-	def exit_error():
-		sys.exit(-1)
+    def exit_fail():
+        sys.exit(1)
+
+    def exit_error():
+        sys.exit(-1)
 else:
-	from preupg.script_api import *
+    from preupg.script_api import *
 
 #END GENERATED SECTION
 # exit functions are exit_{pass,not_applicable, fixed, fail, etc.}
@@ -33,8 +33,8 @@ else:
 # for logging in-place risk use functions log_{extreme, high, medium, slight}_risk
 
 def ifcfg_error():
-	"""https://bugzilla.redhat.com/show_bug.cgi?id=1111174
-1) kernel names (eth0, eth1, ...) 
+    """https://bugzilla.redhat.com/show_bug.cgi?id=1111174
+1) kernel names (eth0, eth1, ...)
 a) DEVICE + HWADDR specified in ifcfg files = not safe, udev in rhel7 is not able to swap names (onlky exception here are machines with only one network interface, they should be fine) <- this is only thing which is covered by content which I wrote
 b) only DEVICE specified in ifcfg files = not safe, udev will probably name the interface differently in rhel7
 c) only HWADRR specified in ifcfg files = uh, device will be set up correctly according to ifcfg file but name could be different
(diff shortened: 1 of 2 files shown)


RHEL6_7/system/initscripts/ifcfg/check
```diff
@@ -51,99 +51,115 @@ I case 3, to be completely correct you could check that user is not using rhel7
...
+    def is_udev(name):
+        if name[0:2] in ("en", "sl", "wl", "ww"):
+            if name[2] in ("b", "c", "o", "s", "x", "P", "p"):
+                return True
+        return False
+
+    def fix_ifnames():
+        script_result = subprocess.Popen("./preupg_network.sh", shell=True)
+        script_result.communicate()
+        exit_code = script_result.wait()
```
bocekm: use just `script_result.returncode` instead of `script_result.wait()`. The `wait()` is already being called internally by `communicate()`.

RHEL6_7/system/initscripts/ifcfg/check
```diff
@@ -51,99 +51,115 @@ I case 3, to be completely correct you could check that user is not using rhel7
...
+        if exit_code == 1:
+            log_extreme_risk("The network interface names on the source system are in inconsistent state. Upgrade is not possible")
+        elif exit_code == 2:
+            log_extreme_risk("MACADDR directive found in the ifcfg files. Comment it out and rerun preupgrade-assistant Upgrade is not possible")
+        return True
```
bocekm: You can use just `return` or not call return at all. Having `return True` may be misleading to some because it indicates you return some specific value even though:
1) nobody consumes it
2) there is no other return, that would return something else, for example False

RHEL6_7/system/initscripts/ifcfg/preupg_network.sh
```diff
@@ -0,0 +1,54 @@
...
+        fi
+    fi
+
+    echo -e "SUBSYSTEM==\"net\", ACTION==\"add\", DRIVERS==\"?*\", $mac_rule, $dev_type_rule, KERNEL==\"${orig_name_base}*\", NAME=\"$new_full_name\"\n" >> udev_temp
+
+done
+
+mv udev_temp "$temp_conf_dir"
+
+echo "To workaround the problem with interfaces renaming, you can replace ifcfg-eth* files in $conf_dir with ifcfg-net* files from $temp_conf_dir \
```
bocekm: The solution states:
> It is highly recommended to rename such interfaces or remove the DEVICE line from the ifcfg configuration files.

I would word the suggestion here in a bit different way, because the sentence above does not sound like a problem to be workarounded. What about:
"We have generated ifcfg files and udev rules that will help you to rename the ethX network interfaces to netX. Perform the following steps to use netX as a new name for the ethX interfaces:"

bocekm (dismissed):

AloisMahdal:
Sorry for taking so long to respond, @jmazanek.  I have few questions:

 1. How are the `ifcfg-preupg-*` files named?  Is it one file per device, or other way?
 2. Is this really only devices named `eth?`
 3. What if there is only one device? (named `eth?`)

Also, I believe that the convention is that files in dirtyconf/cleanconf should already follow path they will be delivered to. I.e. a file that should become `/etc/sysconfig/network-scripts/ifcfg-foo` should be `...dirtyconf/etc/sysconfig/network-scripts/ifcfg-foo`.  (This allows comparing the files using tools like meld or kdiff3.)  So my suggestion would be to change the paths.

bocekm:
> What if there is only one device? (named eth?)

This use case has been covered even before this fix - when there's one eth device, there's no problem so no action needed. See https://github.com/upgrades-migrations/preupgrade-assistant-modules/blob/99147cfa52918135cb80c350d6967f554b7ed45d/RHEL6_7/system/initscripts/ifcfg/check#L137

---
I would agree with changing the dirtyconf directory, however Jakub is going to update the module to produce preupgrade script that will be executed by the redhat-upgrade-tool before rebooting into the initramdisk.

AloisMahdal:
Well, I don't care what is in the code; real serious QE :100: never does :stuck_out_tongue_winking_eye:.  I'm asking for consensus as to what *should* happen.

> [...] when there's one eth device, there's no problem so no action needed [...]

So now I have seen the code I can't unsee it, and it actually contradicts what you said: no action **or** `log_slight_risk` (with a dubious text)?

> [...] however Jakub is going to update the module [...]

I know that there are further changes planned in future, but that does not prevent us from following our conventions.  Or do you mean that he wants to do in this PR?  If so, I'll have to wait with the ack; I can't ack something without knowing what is intended final behavior.

Anyway, still waiting for my Q1-3.

jmazanek:
Thanks @AloisMahdal for looking at it. When you asked your questions, the module was still in incomplete state. So I will explain the behaviour of the updated version.

This fix is an update of the already existing module. Previously the module had been just informative, analysing the network interface naming on the source system and warning the user about the risk of inconsistent network naming state on the upgraded system. This can cause an inaccessible network after upgrade. This issue appears when there are multiple network interfaces present on the system, that use kernel names with "eth" prefix. This fix creates the pre-upgrade script that is executed by redhat-upgrade-tool before an actual upgrade. The following are the cases that module handles:

1. Any of the ifcfg-* files in /etc/sysconfig/network-scripts contains MACADDR directive.

       Result: Fail & inhibits an upgrade and prints a message informing the user.
 
2. Any of the interfaces has rename prefix. (Corner case, hardly reproducible)

       Result: Fail & inhibits an upgrade  and prints a message informing the user.

If none of the above is true, then:

3.  Multiple network interfaces with eth prefix exist on the system.

     a) No network interfaces with net prefix exist on the system.

         Result: Needs_action. Interfaces are renamed with net prefix, while persistently keeping the index    
         number and mac address

     b) There are network interfaces with net prefix on the system 

          Result: Needs_action . Interfaces are renamed with net prefix, keeping persistently a mac address.
          Index number is increased to first available number to prevent conflicts.

4. Only one interface with eth prefix and all other options 
    Result: pass or needs_inspection. This is defined in the original script. It is outside of the scope of this fix
    and does not need to be tested.


How to test

1. Setup multiple network interfaces with eth prefix. ( set some of them as a slave interfaces to bond, 
    configure some to use static IP and some to use dynamic IP)

2.  Run preupgrade assistant and examine result

3. Upgrade the system using redhat-upgrade-tool

Interface names should be persistently set to net prefix and network should be accessible after upgrade.

jmazanek:
One more note. Regarding the name of the directory with fixed configs.  The directory contains not only fixed ifcfg-* files but also file with udev rules. They belong to the different directories on the system, so I have chosen a unique name.  I can change it, if it is a big problem.

RHEL6_7/system/initscripts/ifcfg/check
```diff
@@ -51,99 +51,114 @@ I case 3, to be completely correct you could check that user is not using rhel7
...
+        return False
+
+    def fix_ifnames():
+        script_result = subprocess.Popen("./preupg_network.sh", shell=True)
+        script_result.communicate()
+        exit_code = script_result.returncode
+        if exit_code == 1:
+            log_extreme_risk("The network interface names on the source system are in inconsistent state. Upgrade is not possible")
+        elif exit_code == 2:
+            log_extreme_risk("MACADDR directive found in the ifcfg files. Comment it out and rerun preupgrade-assistant Upgrade is not possible")
```
bocekm: "Comment it out and rerun preupgrade-assistant Upgrade is not possible"
->
"Comment it out and rerun the Preupgrade Assistant. Without that upgrade is not possible."

RHEL6_7/system/initscripts/ifcfg/check
```diff
@@ -51,99 +51,114 @@ I case 3, to be completely correct you could check that user is not using rhel7
...
+                    ethx_with_addr_count +=1
+                else:
+                    ethx_without_addr_count +=1
+                    log_slight_risk(full_path + " is old style ethX name without HWADDR, its name can change after the upgrade.")
+                    warning = True
+            elif is_udev(name):
+                log_slight_risk(full_path + " variable DEVICE is similar to udev predictable network naming scheme. It might cause conflicts.")
+                warning = True
+
+        if 1 in (ethx_with_addr_count, ethx_without_addr_count) > 1:
```
bocekm: What do you want this condition to do? I think there's a 99% chance this construct does not do what you want. Or at least it is very confusing.

I checked the precedence of operators in Python and the `in` is on the same level of precedence as the `>` operator. Then it behavies similar to this example:  x < y <= z is equivalent to x < y and y <= z.

Thus:
`1 in (ethx_with_addr_count, ethx_without_addr_count)` returns `True` (which in Python is the same as `1`) if either of the two variables holds `1`, and then you logically AND this result with `(ethx_with_addr_count, ethx_without_addr_count) > 1` comparison which I'm not sure how it should behave. IMO, this is not easily readable and thus clean code.

bocekm (changes requested):

RHEL6_7/system/initscripts/ifcfg/check
bocekm: It seems you just want to know if either x or y has value bigger than 1, so I would write:
`if ethx_with_addr_count > 1 or ethx_without_addr_count > 1:`

RHEL6_7/system/initscripts/ifcfg/check
```diff
@@ -7,33 +7,33 @@ import re
...
 #END GENERATED SECTION
 # exit functions are exit_{pass,not_applicable, fixed, fail, etc.}
 # logging functions are log_{error, warning, info, etc.}
 # for logging in-place risk use functions log_{extreme, high, medium, slight}_risk
 
 def ifcfg_error():
-	"""https://bugzilla.redhat.com/show_bug.cgi?id=1111174
+    """https://bugzilla.redhat.com/show_bug.cgi?id=1111174
 1) kernel names (eth0, eth1, ...) 
 a) DEVICE + HWADDR specified in ifcfg files = not safe, udev in rhel7 is not able to swap names (onlky exception here are machines with only one network interface, they should be fine) <- this is only thing which is covered by content which I wrote
```
bocekm: Update these 1) a) b) c) points and references to them a few lines below/

bocekm (dismissed):

RHEL6_7/system/initscripts/ifcfg/preupg_network.sh
```diff
@@ -0,0 +1,65 @@
...
+        fi
+    fi
+
+    echo -e "SUBSYSTEM==\"net\", ACTION==\"add\", DRIVERS==\"?*\", $mac_rule, $dev_type_rule, KERNEL==\"${orig_name_base}*\", NAME=\"$new_full_name\"\n" >> udev_temp
+
+done
+
+
+mv udev_temp "$temp_conf_dir"
+echo "cp -p $temp_conf_dir/udev_temp /etc/udev/rules.d/70-persistent-net.rules" >> "$preupg_script"
```
bocekm, line 85: The cp will overwrite the content of /etc/udev/rules.d/70-persistent-net.rules by the content of udev_temp. Is it expected? Shouldn't it append instead?

jmazanek: No, this is an intended behaviour

RHEL6_7/system/initscripts/ifcfg/check
jmazanek: Thanks for the review on this. I have corrected the statement in a new build

RHEL6_7/system/initscripts/ifcfg/check
jmazanek: Fixed

tcerna:
Hi, I did first check of result.html file and found two problems:
- message "preupg.risk.HIGH: You use multiple network devices with old style 'ethX' names." is written several times according to number of eth interfaces. there should be only one risk about it, others are useless.
- typo in remediation description:
  - old: The interfaces with "eth" prefix will be renamed to use "**eth**" prefix by...
  - new: The interfaces with "eth" prefix will be renamed to use "**net**" prefix by...

jmazanek:
@tcerna  thanks for spotting it. This is now fixed with a new build

bocekm (dismissed):

pirat89:
Original commit has been split into two commits:
- refactoring commit (contains changes only about whitespaces)
- solution itself

Against previous solution, I removed trailing whitespaces in the refactoring commit. Everything else is same. (added back the pre-upgrade script :-) Thanks @bocekm for catch :-) )

bocekm (approved):
Re-stamping after the split of one commit to two for clarity.

RHEL6_7/system/initscripts/ifcfg/preupg_network.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
+
+orig_name_base="eth"
+new_name_base="net"
+conf_dir='/etc/sysconfig/network-scripts'
+temp_conf_dir='/root/preupgrade/dirtyconf/etc/sysconfig/network-configuration_fixed/'
+preupg_script='/root/preupgrade/preupgrade-scripts/rename_network.sh'
```
pirat89, line 8: @jmazanek: you would be able to use here probably API variables. it would be better use them when it works. Still you can export them if these are not set. But it is not blocker for this PR now.

RHEL6_7/system/initscripts/ifcfg/preupg_network.sh
```diff
@@ -0,0 +1,75 @@
...
+udev_command='/sbin/udevadm info -a  -p'
+dev_path='/sys/class/net'
+index=0
+
+declare -a netX_indices=$(ip a | awk -F': ' '/^[[:digit:]]: net[[:digit:]]+/ {print $2  }' | sed -e "s/net//g" )
```
bocekm: Can it happen that the `ip a` would return the network devices in non-alphabetical order, e.g. net1 first, net0 second. If yes, then it would break looking for the non-collision index in the three nested loops below. Sorting it would fix that.

RHEL6_7/system/initscripts/ifcfg/preupg_network.sh
```diff
@@ -0,0 +1,75 @@
...
+    dev_type_rule=$($udev_command "${dev_path}/${orig_full_name}" | egrep "ATTR\{type\}")
+    index=$(echo $orig_full_name |sed -e "s/^$orig_name_base//g")
+
+    for net_index in ${netX_indices[@]}
+    do
+        while [ $index -eq $net_index ]
+        do
+            index=$(expr $index + 1)
+        done
+        netX_indices=(${netX_indices[@]} $index )
```
bocekm: This loop within a loop within a loop is so hard to grasp for me. Maybe it's just me, but I needed like an hour to decipher the inner workings of these nested loops.
a) When looking for what `netX_indices=(${netX_indices[@]} $index )` is doing, I found out it's the same as `netX_indices+=($index)`, which seems more readable to me.
b) Note to whoever wondering whether it's going to affect the for loop, if you add something to `netX_indices` within the loop. The answer is no.
c) The b) leads me to this: I don't understand why you add the `index` to the `netX_indices` array within the for loop. I would move it after the for loop.
d) No matter the c), this algorithm wouldn't work for example in case there's eth0, eth1, net0, net2 original setup. The eth0 will get renamed to net1 and eth1 will get renamed to net2, which is however already taken.

bocekm (dismissed):
Changing of the "net" index in case of a conflict means that eth0 may not translate to net0 which is what the user would expect from the description. There's currently no way for the user to find out that our module is going to rename eth0 to for example net3. I see two options:
1) Instead of trying to resolve the conflict, end the script with extreme risk, implement the configurable prefix in the `/etc/preupgrade-assistant.conf` and tell users in the report to use some prefix that has no conflict with their existing network device names.
2) Tell users in the report that the automated script will rename their ethX to netY, no matter if X=Y or X!=Y. For each of their ethX device. 

Rant only: I'd prefer functions with no more 10 lines instead of one do-it-all spaghetti code in the `preupg_network.sh`. But keep it as is now.

RHEL6_7/system/initscripts/ifcfg/preupg_network.sh
jmazanek: yes, theoretically it could happen ( even if it is counter-intuitive to name it like that). I will sort it

RHEL6_7/system/initscripts/ifcfg/preupg_network.sh
jmazanek: a) Seems readable to me, but I don't mind changing it
c) You are right, it was meant to be after loop. It still worked but it was not necessary to run in the loop
b) It does though. You can easily test it in the standalone script like this ( with correction from the point c) )

     #!/bin/bash
     my_array=( 1 2 3)
     your_array=( 1  3)

     for m in ${my_array[@]}
     do
            for y in ${your_array[@]}
            do
                  echo "your_array before is : ${your_array[@]}"
                  while [ $m -eq $y ]
                  do
                          m=$(expr $m + 1 )
                  done
           done
           your_array=( ${your_array[@]} $m )
           echo "your_array after is : ${your_array[@]}"
    done

d) It does work. The `$index` is matched against each `$net_index` and as long as it matches it is increased by `1`

jmazanek:
@bocekm 
I disagree with your proposed solutions. If you think there is any confusion on the user side about the new interface names I will improve the message. But the naming conflict resolution works as expected ( I have elaborated it in the commit comments )and it is IMO useful feature.

jmazanek:
Created new build that just edit the user messages to better reflect current version of the the module ( automating renaming, handling pre-existing netX interfaces). Nothing else besides these messages has been touched.

tcerna:
Verified in preupgrade-assistant-el6toel7-0.7.2-0.201803281712Z.PR68.el6_8.noarch:
  * output in result.html file was checked
  * pre-upgrade scritp rename_network.sh was created and seems well
  * redhat-upgrade-tool works well without tracebacks and fails (thank to Kuba for fixing it)
  * machines boot to rhel-7.5
  * ethX interfaces were renamed to netX and all set interfaces are available (tested via ping, scp and ssh)
  * test that renaming works well (have net0, net1, eth0, eth1 interfaces and check that renaming will be eth0->net2 and eth1->net3)
  * mac addresses for each interface are corresponding to its adress before upgrade 
  * several posibilities were checked and work well

bocekm:
@jhornice, could you please review the user-facing texts of this module?

AloisMahdal:
`qa_ack`: Acking retrospectively; testing oracle and strategy has been discussed above.

RHEL6_7/system/initscripts/ifcfg/check
```diff
@@ -7,143 +7,157 @@ import re
...
+            if name[2] in ("b", "c", "o", "s", "x", "P", "p"):
+                return True
+        return False
+
+    def fix_ifnames():
+        script_result = subprocess.Popen("./preupg_network.sh", shell=True)
+        script_result.communicate()
+        exit_code = script_result.returncode
+        if exit_code == 1:
+            log_extreme_risk("The network interface names on the source system are in inconsistent state. Upgrade is not possible")
```
jhornice: "The network interface names on the source system are in an inconsistent state, so the upgrade is not possible"

RHEL6_7/system/initscripts/ifcfg/check
```diff
@@ -7,143 +7,157 @@ import re
...
+        elif exit_code == 2:
+            log_extreme_risk("MACADDR directive found in the ifcfg files. Comment it out and rerun preupgrade-assistant. Without that upgrade is not possible")
```
jhornice: "Comment out the MACADDR directive in the ifcfg files and run the Preupgrade Assistant again to enable the upgrade."

RHEL6_7/system/initscripts/ifcfg/check
```diff
@@ -7,143 +7,157 @@ import re
...
+        addr = get_variable(full_path, "HWADDR")
+        name = get_variable(full_path, "NAME" if devel_time else "DEVICE")
+        log_debug("checking " + script + ", name: " + str(name) + " hwaddr: " + str(addr))
+        short_name = script[len(ifcfg_prefix):]
+        if short_name == "lo": #loopback
+            continue
+        if not name:
+            warning = True
+            if not addr:
+                log_slight_risk(full_path + " does not have DEVICE nor HWADDR set, what kind of device is it?")
```
jhornice, line 128: full_path + " does not have DEVICE nor HWADDR set, what kind of device is it?"
what are you trying to say here?

RHEL6_7/system/initscripts/ifcfg/check
```diff
@@ -7,143 +7,157 @@ import re
...
+            else:
+                log_slight_risk(full_path + " does not have DEVICE set, its name can change after the upgrade.")
```
jhornice, line 130: full_path + " does not have DEVICE set, its name can change after the upgrade."
the same as above+what does "its name can change" actually mean?

RHEL6_7/system/initscripts/ifcfg/preupg_network.sh
```diff
@@ -0,0 +1,92 @@
...
+    for net_index in ${netX_indices[@]}
+    do
+        while [ $index -eq $net_index ]
+        do
+            index=$(expr $index + 1)
+        done
+    done
+    netX_indices=(${netX_indices[@]} $index )
+    new_full_name="${new_name_base}$index"
+    log_info "On the upgraded system the $orig_full_name interface will become $new_full_name"
```
jhornice: "On the target system, the $orig_full_name interface will become $new_full_name."

RHEL6_7/system/initscripts/ifcfg/preupg_network.sh
```diff
@@ -0,0 +1,92 @@
...
+done
+
+
+mv udev_temp "$temp_conf_dir"
+echo "cp -p $temp_conf_dir/udev_temp /etc/udev/rules.d/70-persistent-net.rules" >> "$preupg_script"
+echo 'service network start'  >> "$preupg_script"
+chmod u+x "$preupg_script"
+
+
+echo "The interfaces with \"$orig_name_base\" prefix will be renamed to use \"$new_name_base\" prefix by $preupg_script script executed by redhat-upgrade-tool before the upgrade.
```
jhornice, line 90: I'm a bit lost here. "The interfaces with \"$orig_name_base\" prefixes will be renamed to use \"$new_name_base\" prefixes before the upgrade." Shouldn't it be enough?

RHEL6_7/system/initscripts/ifcfg/preupg_network.sh
```diff
@@ -0,0 +1,92 @@
...
+      This process includes the restart of the network" >> solution.txt
```
jhornice, line 91: It includes? It needs?

RHEL6_7/system/initscripts/ifcfg/check
```diff
@@ -124,18 +133,23 @@ I case 3, to be completely correct you could check that user is not using rhel7
                 if addr:
                     ethx_with_addr_count +=1
                 else:
+                    ethx_without_addr_count +=1
```
jhornice: its name CAN change?

RHEL6_7/system/initscripts/ifcfg/check
```diff
@@ -124,18 +133,23 @@ I case 3, to be completely correct you could check that user is not using rhel7
...
                     log_slight_risk(full_path + " is old style ethX name without HWADDR, its name can change after the upgrade.")
```
jhornice: its name CAN change? or WILL change?

jhornice: its name CAN change or WILL change?

pirat89 (dismissed):
It looks good, just with several things that can be done next milestone and are not blocking this PR because of lack of time. See the issue #72

RHEL6_7/system/initscripts/ifcfg/check
jmazanek: will fix

RHEL6_7/system/initscripts/ifcfg/check
jmazanek: will fix

RHEL6_7/system/initscripts/ifcfg/check
jmazanek: I'm not sure. This is a message from the original code. We can address it in a batch after the current release

RHEL6_7/system/initscripts/ifcfg/check
jmazanek: Basically ( just like in the above case) HWADDR and DEVICE directives are normally used by the kernel to determine the mapping of the interface names.  The order in which the interfaces come up is not possible to determine and the interface name CAN change.

RHEL6_7/system/initscripts/ifcfg/preupg_network.sh
jmazanek: will fix

RHEL6_7/system/initscripts/ifcfg/preupg_network.sh
jmazanek: It could be enough, but I think it is still useful information for the user to see which specific script does the change, so it can be edited if necessary

RHEL6_7/system/initscripts/ifcfg/preupg_network.sh
jmazanek: I think includes is better ( we inform the user, that the network will be restarted as a part of the interface renaming process). You can rename the interfaces without restart, but not persistently ( just in a runtime), so it is not strictly needed, but it is used by p-a.

RHEL6_7/system/initscripts/ifcfg/check
jmazanek: yes CAN is correct (explained above)

jmazanek:
@jhornice , thanks for the text review, it was included in the new build
@pirat89 , thanks for the issue, we will address it later

RHEL6_7/system/initscripts/ifcfg/check
jhornice: ...has neither DEVICE nor HWADDR set...

pirat89: appended to the related issue.

pirat89 (approved):

RHEL6_7/system/initscripts/ifcfg/check
bocekm: I'd use MAY, but I'm a bit late to the party :)

jhornice: I'd use MIGHT, but I didn't:(

PR #69, opened 2018-03-05
Unset the release

bertinatto:
This way, after the upgrade the subscription-manager always downloads updates for the latest minor RHEL 7 version.

This action covers the corner case of the RHEL 6 system having a release set (e.g. `subscription-manager release --set=6.9`). That would cause the upgrade to fail as yum would be trying to access RHEL 6.9 release under RHEL 7 repos.

### DIFF
--- a/RHEL6_7/system/subscription-manager/module.ini
+++ b/RHEL6_7/system/subscription-manager/module.ini
@@ -1,5 +1,5 @@
 [preupgrade]
 content_title: Red Hat Subscription Manager
-content_description: The module adds a new option 'full_refresh_on_yum' to the configuration file, if it is not there already.
+content_description: The module adds a new option 'full_refresh_on_yum' to the configuration file, if it is not there already. Also, unsets the release with subscription-manager.
 author: Jakub Dornak <jdornak@redhat.com>
 applies_to: subscription-manager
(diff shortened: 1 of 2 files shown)


bocekm (dismissed):
Please update the module.ini so the module description covers this new action.

jmazanek (approved):
looks good to me

mfranczy (approved):

bocekm:
@jmazanek, @AloisMahdal, I think this could be covered by tests. It looks easy to me to create a test for that:
- set the release before the upgrade,
- perform the upgrade,
- see that the release is not set on the upgraded system.

AloisMahdal:
The thing is, automated upgrade tests are relatively expensive, so I'm a bit reluctant to create one unless it's useful ie. tests probable setup of a component.  Fabio said it's corner case.  So how much in a corner this case is?  I can't say as I have no knowledge about subscription-manager.  Is it valid for user to do that?

Anyway, I will test this test manually, if it's so trivial.

AloisMahdal:
> Anyway, I will test this test manually, if it's so trivial.

Well, it looks like one has to be registered:

    # subscription-manager release --set=6.9; echo $?
    This system is not yet registered. Try 'subscription-manager register --help' for more information.
    1
    # 

and I don't have testing credentials, so it's not so straightforward.

bocekm:
> The thing is, automated upgrade tests are relatively expensive

Argh, we need more QE capacity :) I'll try negotiate another budget to get you a sidekick.

> I don't have testing credentials, so it's not so straightforward.

You used to have QE credentials. You don't anymore? Then yeah, that's a blocker. But for manual test only, you can use your regular employee credentials. If you haven't set it up, I can show you how to get them.

It's true this is a corner case. Nevertheless, I'd like to have even corner cases tested, especially when we have identified them.

PR #73, opened 2018-04-04
Bump to version 0.7.3

pirat89:
- Unset the release when it is set in rhsm
- add quotes for readlink in drivers/* modules so they will not crash
- initscripts/ifcfg: persistant interfaces renaming
- openssh/sshd: the module has been rewritten significantly
  provide correct data to user and be sure the sshd configuration
  is valid


AloisMahdal (approved):
LGTM

PR #75, opened 2018-04-19
RemovedPackages: added noauto postupgrade script

pirat89:
In case user doesn't use --cleanup-post option in r-u-t, there could be kept broken el6 packages which could inhibit user to update their package set. This script can be run by user manually to resolve such issues, keeping non-native packages untouched.

### DIFF
--- a/RHEL6_7/packages/RemovedPackages/check
+++ b/RHEL6_7/packages/RemovedPackages/check
@@ -39,13 +39,13 @@ cp "$POSTUPGRADE_DIR/$(basename $VALUE_RPM_RHSIGNED)" "$NOAUTO_POST_DIR"
 found=0
 rm -f solution.txt
 echo \
-"Some packages were removed between Red Hat Enterprise Linux 6 and Red Hat Enterprise Linux 7. This might break
+"Some packages were removed between Red Hat Enterprise Linux (RHEL) 6 and RHEL 7. This might break
 the upgrade for some of your packages. We are not aware of any compatible
 replacement for these packages.
 
 Use the '--cleanup-post' option of redhat-upgrade-tool to remove them
 automatically during upgrade. You can also run the $NOAUTO_POST_DIR/$NOAUTO_SCRIPT_NAME script
-manually after the upgrade to remove only those Red Hat-signed (RHEL) 6 packages
+manually after the upgrade to remove only those RHEL 6 packages signed by Red Hat
 that have broken dependencies. The script also fixes possible broken
 dependencies of the newly installed packages.
 
@@ -56,7 +56,7 @@ while read pkg
 do
   #skip non-rh and unavailable packages
   is_pkg_installed "$pkg" && is_dist_native $pkg || continue
-  j=" (required by NonRH signed package(s):"
+  j=" (required by packages not signed by Red Hat:"
   for k in $(rpm -q --whatrequires $pkg | grep -v "^no package requires" | \
    rev | cut -d'-' -f3- | rev)
   do
@@ -64,8 +64,8 @@ do
     is_dist_native $k ||  j="$j$k "
   done
   j="${j% })"
-  [ "$j" == " (required by NonRH signed package(s):)" ] && j=""
-  [ -n "$j" ] && log_high_risk "The $pkg $j package was removed between Red Hat Enterprise Linux 6 and Red Hat Enterprise Linux 7."
+  [ "$j" == " (required by packages not signed by Red Hat:)" ] && j=""
+  [ -n "$j" ] && log_high_risk "The $pkg $j package was removed between RHEL 6 and RHEL 7."
   echo "$pkg$j" >>solution.txt
   found=1
 done < "$RemovedPkgs"
@@ -76,21 +76,21 @@ grep -v required solution.txt | grep -v " " | grep -v "^$" >> "$KICKSTART_DIR/Re
 grep required "$KICKSTART_DIR/RemovedPkg-required" >/dev/null || rm "$KICKSTART_DIR/RemovedPkg-required"
 grep [a-zA-Z] "$KICKSTART_DIR/RemovedPkg-optional" >/dev/null || rm "$KICKSTART_DIR/RemovedPkg-optional"
 [ -f "$KICKSTART_DIR/RemovedPkg-required" ] && \
-  echo " * RemovedPkg-required - This file contains all Red Hat Enterprise Linux 6 packages that were removed in Red Hat Enterprise Linux 7. There is no known compatible-enough alternative for them. As some of your packages depend on them, check the changes carefully." >>"$KICKSTART_README"
+  echo " * RemovedPkg-required - This file contains all RHEL 6 packages that were removed in RHEL 7. There is no known compatible alternative for them. As some of your packages depend on them, check the changes carefully." >>"$KICKSTART_README"
 [ -f "$KICKSTART_DIR/RemovedPkg-optional" ] && \
   echo " * RemovedPkg-optional - Similar to the RemovedPkg-required file, but in this case, no package not signed by Red Hat requires this. It is more of an informational thing for you, so that you can deal with the unavailability of these packages." >>"$KICKSTART_DIR/README"
 
 echo \
 "
-If a package not signed by Red Hat requires these packages, you may need to ask your
-vendor to provide an alternative solution, or you may get the missing package from
-other sources than Red Hat Enterprise Linux.
+If a package not signed by Red Hat requires these packages, ask your
+vendor to provide an alternative solution, or get the missing package from
+other sources than RHEL.
 
 " >>solution.txt
-[ $found -eq 1 ] && log_high_risk "After upgrading to Red Hat Enterprise Linux (RHEL) 7, there are still some RHEL 6 packages left. Add the '--cleanup-post' option to redhat-upgrade-tool to remove them automatically. You can also run the $NOAUTO_POST_DIR/$NOAUTO_SCRIPT_NAME script manually after the upgrade to remove only those Red Hat-signed RHEL 6 packages that have broken dependencies."
+[ $found -eq 1 ] && log_high_risk "After upgrading to RHEL 7, there are still some RHEL 6 packages left. Add the '--cleanup-post' option to redhat-upgrade-tool to remove them automatically. You can also run the $NOAUTO_POST_DIR/$NOAUTO_SCRIPT_NAME script manually after the upgrade to remove only those Red Hat-signed RHEL 6 packages that have broken dependencies."
 
 [ $found -eq 1 ] && log_medium_risk "\
-Some packages installed on the system were removed between Red Hat Enterprise Linux 6 and Red Hat Enterprise Linux 7. This might break the functionality of the packages that depend on the removed packages." && exit $RESULT_FAIL
+Some packages installed on the system were removed between RHEL 6 and RHEL 7. This might break the functionality of the packages that depend on the removed packages." && exit $RESULT_FAIL
 
 rm -f solution.txt && touch solution.txt
 
--- a/RHEL6_7/packages/RemovedPackages/check
+++ b/RHEL6_7/packages/RemovedPackages/check
@@ -44,10 +44,10 @@ the upgrade for some of your packages. We are not aware of any compatible
 replacement for these packages.
 
 Use the '--cleanup-post' option of redhat-upgrade-tool to remove them
-automatically during upgrade. You can also run manually the $NOAUTO_POST_DIR/$NOAUTO_SCRIPT_NAME script
-after the upgrade to remove only the original native packages with broken
-dependencies. The script also fixes possible broken dependencies of the newly
-installed packages.
+automatically during upgrade. You can also run the $NOAUTO_POST_DIR/$NOAUTO_SCRIPT_NAME script
+manually after the upgrade to remove only those Red Hat-signed (RHEL) 6 packages
+that have broken dependencies. The script also fixes possible broken
+dependencies of the newly installed packages.
 
 The following packages are no longer available:" >solution.txt
 
@@ -87,7 +87,7 @@ vendor to provide an alternative solution, or you may get the missing package fr
 other sources than Red Hat Enterprise Linux.
 
 " >>solution.txt
-[ $found -eq 1 ] && log_high_risk "After upgrading to Red Hat Enterprise Linux 7, there are still some el6 packages left. Add the '--cleanup-post' option to redhat-upgrade-tool to remove them automatically. You can also run manually the $NOAUTO_POST_DIR/$NOAUTO_SCRIPT_NAME script after the upgrade to remove only old native packages with broken dependencies."
+[ $found -eq 1 ] && log_high_risk "After upgrading to Red Hat Enterprise Linux (RHEL) 7, there are still some RHEL 6 packages left. Add the '--cleanup-post' option to redhat-upgrade-tool to remove them automatically. You can also run the $NOAUTO_POST_DIR/$NOAUTO_SCRIPT_NAME script manually after the upgrade to remove only those Red Hat-signed RHEL 6 packages that have broken dependencies."
 
 [ $found -eq 1 ] && log_medium_risk "\
 Some packages installed on the system were removed between Red Hat Enterprise Linux 6 and Red Hat Enterprise Linux 7. This might break the functionality of the packages that depend on the removed packages." && exit $RESULT_FAIL
--- a/RHEL6_7/packages/RemovedPackages/check
+++ b/RHEL6_7/packages/RemovedPackages/check
@@ -17,13 +17,24 @@ cat "$COMMON_DIR"/default*_removed* | grep -v "\.so" | cut -f1 -d' ' | sort | un
   exit $RESULT_ERROR
 
 SCRIPT_NAME="postupgrade_cleanup.sh"
+NOAUTO_SCRIPT_NAME="native_rpm_dep_fix.sh"
 POSTUPGRADE_DIR="$VALUE_TMP_PREUPGRADE/postupgrade.d/clean_rhel6_pkgs"
+NOAUTO_POST_DIR="$VALUE_TMP_PREUPGRADE/noauto_postupgrade.d/rpmfix"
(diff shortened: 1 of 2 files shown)


pirat89:
@AloisMahdal In this case, I think just partial tests are required. I guess we do not have tests for this module specifically, do we? Currenly expected behaviour:
- expected files are copied under the noauto_postupgrade.d dir    
- the script remove all old native packages with broken dependencies    
- custom rpms are by default kept untouched (doesn't matter whether deps are broken or not here)
  - when the script is called with parameter `--rm-all`, remove even non-native packages when they have broken deps
- in case some new-native packages have broken dependencies, install dependencies when they are available
- in case something is wrong, inform user that manual steps are required       
     
- in case a "branch" of tree with broken dependencies is longer than 10 rpms, probably some packages have not been still removed and script has to be run again; in that case, user is informed whether some rpms with broken deps remain and should be removed.
  - Example: let's have 15 rpms with cascade deps (rpm15 requires rpm14, ...) and only rpm1 has now broken dependency. By removing of rpm1 we break deps of rpm2 (nothing else). Remove rpm2 in another iteration so rpm3 has broken deps, ....

jmazanek (dismissed):
Reviewed. Seems good to me

RHEL6_7/packages/RemovedPackages/noauto_postupgrade.d/native_rpm_dep_fix.sh
```diff
@@ -22,22 +29,33 @@ is_dist_native() {
 }
 
 rm_native_broken_old_rpms() {
+    #
+    # Remove native RPMs with broken dependencies.
+    #
+    # The fce can be affected by the $RM_CUSTOM_RPM variable to remove even
```
AloisMahdal: "fce" is a contraction of Czech "funkce".

pirat89: Thanks :-)

pirat89: fixed

pirat89:

pirat89:
--test--

RHEL6_7/packages/RemovedPackages/noauto_postupgrade.d/native_rpm_dep_fix.sh
```diff
@@ -99,7 +103,7 @@ for counter in {0..10}; do
     # try just limited number of loops in most to be sure that scripts ends
     # always (in reasonable time)
     # .. in case that nothing is removed (returned 0) break the loop
-    rm_native_broken_old_rpms && break
+    rm_broken_old_rpms "$RM_CUSTOM_RPM" && break
     [ $counter -eq 10 ] && {
         [ -z "$(get_broken_deps_list)" ] && {
             log_warning "Reached iteration limit but some RPMs are still broken."
```
jmazanek, line 109: The log warning does not tell the user, how to resolve the situation. Add there the information about, what should user do, when it happens (re-run the script).

RHEL6_7/packages/RemovedPackages/check
```diff
@@ -68,8 +85,9 @@ echo \
 If a package not signed by Red Hat requires these packages, you may need to ask your
 vendor to provide an alternative solution, or you may get the missing package from
 other sources than Red Hat Enterprise Linux.
+
 " >>solution.txt
-[ $found -eq 1 ] && log_high_risk "After upgrading to Red Hat Enterprise Linux 7, there are still some el6 packages left. Add the '--cleanup-post' option to redhat-upgrade-tool to remove them automatically."
+[ $found -eq 1 ] && log_high_risk "After upgrading to Red Hat Enterprise Linux 7, there are still some el6 packages left. Add the '--cleanup-post' option to redhat-upgrade-tool to remove them automatically. You can also run manually the $NOAUTO_POST_DIR/$NOAUTO_SCRIPT_NAME script after the upgrade to remove only old native packages with broken dependencies."
```
bocekm: - *Red Hat Enterprise Linux 7, there are still some el6* **->** *Red Hat Enterprise Linux (RHEL) 7, there are still some RHEL 6*
- *run manually the XY script* **->** *run the XY script manually*
- *only old native packages with broken dependencies* **->** *only those Red Hat-signed RHEL 6 packages that have broken dependencies*

RHEL6_7/packages/RemovedPackages/check
```diff
@@ -32,6 +43,12 @@ echo \
 the upgrade for some of your packages. We are not aware of any compatible
 replacement for these packages.
 
+Use the '--cleanup-post' option of redhat-upgrade-tool to remove them
+automatically during upgrade. You can also run manually the $NOAUTO_POST_DIR/$NOAUTO_SCRIPT_NAME script
+after the upgrade to remove only the original native packages with broken
```
bocekm: - *run manually the XY script* **->** *run the XY script manually*
- *only old native packages with broken dependencies* **->** *only those Red Hat-signed RHEL 6 packages that have broken dependencies*

RHEL6_7/packages/RemovedPackages/noauto_postupgrade.d/native_rpm_dep_fix.sh
```diff
@@ -0,0 +1,135 @@
...
+
+    get_broken_deps_list | grep "\.el6" > rpms_broken
+    local REMOVED_FLAG=0
+    local line=""
+    while IFS= read -r line || [ -n "$line" ]; do
+        NAME=$(rpm -q --qf '%{NAME}' $line)
+        is_dist_native "$NAME" || [ $rm_custom_enabled -eq 1 ] || {
+            # it's not dist native package -> skip it
+            echo "    $NAME" >> "$kept_broken_rpms"
+            log_info_verbose "Skip the non-native $line RPM."
```
bocekm: _non-native_ **->** *3rd party or custom*

RHEL6_7/packages/RemovedPackages/noauto_postupgrade.d/native_rpm_dep_fix.sh
```diff
@@ -0,0 +1,135 @@
...
+    case $1 in
+        -a | --rm-all)
+            RM_CUSTOM_RPM=1
+            ;;
+        -v | --verbose)
+            VERBOSE=1
+            ;;
+        -h | --help)
+            echo "USAGE: native_rpm_dep_fix [-a|--rm-all] [-v|--verbose] [-h|--help]"
+            echo "    -a | --rm-all   Enable remove of non-native RPMs"
```
bocekm: _remove of non-native_ **->** *removal of 3rd party or custom*

RHEL6_7/packages/RemovedPackages/noauto_postupgrade.d/native_rpm_dep_fix.sh
```diff
@@ -0,0 +1,135 @@
...
+    shift
+done
+
+
+cd $(dirname "$0")
+rm -f "$kept_broken_rpms"
+touch "$kept_broken_rpms"
+
+[ $RM_CUSTOM_RPM -eq 1 ] \
+    && log_info_verbose "Remove of non-native conflicting RPMs has been enabled."
```
bocekm: _Remove of non-native_ **->** ***Removal** of 3rd party or custom*
BTW, you're using "conflicting" here for the first and only time. I would use what you use on the other places - "with broken dependencies".

RHEL6_7/packages/RemovedPackages/noauto_postupgrade.d/native_rpm_dep_fix.sh
```diff
@@ -0,0 +1,135 @@
...
+            log_info "You can run the application again to try to remove rest of RPMs."
+            FAIL=1
+        }
+    }
+done
+
+broken_nonnative_rpms=$(wc -l <"$kept_broken_rpms")
+[ $broken_nonnative_rpms -gt 0 ] && {
+    # this is obviously irrelevant in case of use the -a option
+    log_warning "Detected several non-native RPMs with broken dependencies:"
```
bocekm: *non-native* **->** *3rd party or custom RHEL 6*

bocekm (changes requested):
I'd like to avoid the term *native* - it's not used in any other module (in a user-facing messages) and it's not a widespread term so i think it would be causing confusion for customers.
I'd rename the *native_rpm_dep_fix* script to something like *rm_broken_el6_rpms*, especially when it's able to remove also custom and 3rd party packages.
Other solution can be:
- Define the term in the Preupgrade Assistant docs
- Start the solution text with definition of the term

RHEL6_7/packages/RemovedPackages/noauto_postupgrade.d/native_rpm_dep_fix.sh
pirat89: it is written just line below :)

RHEL6_7/packages/RemovedPackages/check
pirat89: fixed

RHEL6_7/packages/RemovedPackages/check
pirat89: fixed

RHEL6_7/packages/RemovedPackages/noauto_postupgrade.d/native_rpm_dep_fix.sh
pirat89: fixed

RHEL6_7/packages/RemovedPackages/noauto_postupgrade.d/native_rpm_dep_fix.sh
pirat89: fixed

RHEL6_7/packages/RemovedPackages/noauto_postupgrade.d/native_rpm_dep_fix.sh
pirat89: You are right. Fixed.

RHEL6_7/packages/RemovedPackages/noauto_postupgrade.d/native_rpm_dep_fix.sh
pirat89: fixed

AloisMahdal:
There was a discussion about this in Bugzilla and/or RH mailing list long time ago.  The "native" term was intentionally introduced to distinguish packages that are owned by distribution from others.  I also recommended to avoid term "3rd party" because it IMHO can be confusing if the "3rd party software" is actually one written by customer--so there's no 3rd party; there's just distribution SW and user SW.  Note that this *is* the only distinction that can ever matter to us: we are'nt supposed to care/imply who maintains the packages if it's not *us*.

Whether `el`, `rhel` or `el6` should be used here depends on whether or not the module is RHEL(6)-specific.  If it inherently is, then it's OK to hardcode `el6` in the name, if not, it's IMO best to avoid.

bocekm:
@AloisMahdal that is nice that you guys introduced it, but it's not documented anywhere and customers will be left wondering what exactly that means. So as I mentioned in the review, I would either rephrase it to avoid the term or explain the term. I agree with you that when using 3rdparty and custom, it's again using terms that are not clearly defined in our project but in my eyes are more self-explanatory. Anyway, I'll reach out to the docs team to ask them what way to go.

bocekm (dismissed):

AloisMahdal:
Sorry, but we won't make it in this release; there's been too many other issues.  I talked to @pirat89 and we've agreed we can move it to the next  one.

RHEL6_7/packages/RemovedPackages/noauto_postupgrade.d/native_rpm_dep_fix.sh
```diff
@@ -0,0 +1,135 @@
...
+#RHSIGNED_PKGS="/var/cache/preupgrade/common/rpm_rhsigned.log"
+RHSIGNED_PKGS="rpm_rhsigned.log"
+kept_broken_rpms="kept_broken_rpms"
+VERBOSE=0
+RM_CUSTOM_RPM=0
+FAIL=0
+
+log_error()   { echo >&2 "Error:   $@"; }
+log_warning() { echo >&2 "Warning: $@"; }
+log_info()    { echo >&2 "Info:    $@"; }
```
AloisMahdal: Mixing array and string.  Did you mean this?

    log_error() { echo >&2 "Error: " "$@"; }

or (argument is line):

    log_error() { for msg in "$@"; do echo >&2 "Error: $msg"; done; }

or just ignore from $2 on::

    log_error() { echo >&2 "Error: $2"; }

RHEL6_7/packages/RemovedPackages/noauto_postupgrade.d/native_rpm_dep_fix.sh
```diff
@@ -0,0 +1,135 @@
+#!/bin/bash
+
+#RHSIGNED_PKGS="/var/cache/preupgrade/common/rpm_rhsigned.log"
+RHSIGNED_PKGS="rpm_rhsigned.log"
+kept_broken_rpms="kept_broken_rpms"
```
AloisMahdal, line 5: I'm not sure why you need variable if the value is always just the variable name itself...

RHEL6_7/packages/RemovedPackages/noauto_postupgrade.d/native_rpm_dep_fix.sh
```diff
@@ -0,0 +1,135 @@
...
+            FAIL=1
+        }
+    }
+done
+
+broken_nonnative_rpms=$(wc -l <"$kept_broken_rpms")
+[ $broken_nonnative_rpms -gt 0 ] && {
+    # this is obviously irrelevant in case of use the -a option
+    log_warning "Detected several 3rd party or custom RHEL 6 RPMs with broken dependencies:"
+    cat "$kept_broken_rpms" | sort | uniq >&2
```
AloisMahdal, line 120: Useless use of cat :)

    <"$stuff" sort | uniq >&2

AloisMahdal (commented):
Just few non-blocking comments.

RHEL6_7/packages/RemovedPackages/noauto_postupgrade.d/native_rpm_dep_fix.sh
```diff
@@ -0,0 +1,135 @@
...
+while [[ -n "$1" ]]; do
+    case $1 in
+        -a | --rm-all)
+            RM_CUSTOM_RPM=1
+            ;;
+        -v | --verbose)
+            VERBOSE=1
+            ;;
+        -h | --help)
+            echo "USAGE: native_rpm_dep_fix [-a|--rm-all] [-v|--verbose] [-h|--help]"
```
AloisMahdal: `native_rpm_dep_fix` is not true anymore.  You could use `${0##*/}` here, I think.

AloisMahdal:
Checked few basic cases and the script works fine.  However, we don't have time anymore for proper verification (which is also why this was not acked.)

RHEL6_7/packages/RemovedPackages/noauto_postupgrade.d/native_rpm_dep_fix.sh
pirat89: I have not been sure that there will not be different name later (because of *broken*) and have not a mood to think about something better. Anyway, yes, it's not neede, but I havn't wanted to correct that several times in case there will be different name used later. And shellckeck works better when you uses variables to detect possible typos :)

RHEL6_7/packages/RemovedPackages/noauto_postupgrade.d/native_rpm_dep_fix.sh
pirat89: I know. But this is only cat I really like, so I put it on many places :heart:

RHEL6_7/packages/RemovedPackages/noauto_postupgrade.d/native_rpm_dep_fix.sh
pirat89: right! will be fixed in the next commit.

RHEL6_7/packages/RemovedPackages/noauto_postupgrade.d/native_rpm_dep_fix.sh
pirat89: The forth option :D :
```
log_error() { echo >&2 "Error: $1"; }
```

pirat89:
@tcerna @AloisMahdal @drehak: Fixed some reported issues and rebased with master. All new changes are part of the last commit. Please, do a review + give an ack (in case you will not find an issue inside the changes :) ) and I will merge it.

tcerna (approved):
Review looks good according to Lojza's comments.

pirat89:
Moved to the next milestone for proper testing.

pirat89:
Removing from the milestone as stil it is not tested.

pirat89:
We will celebrate 1y soon :-P

pirat89:
-- build --

pirat89:
--test--

pirat89:
@bocekm @AloisMahdal what about this *nestor*?

AloisMahdal:
> @bocekm @AloisMahdal what about this _nestor_?

No idea; if it's worth we can try and test it again but it's not like we have extra resources... :-/

pirat89:
Keep in mind it's manual script only - it will not be executed automatically. I guess that manual testing should not take so much time and it could be tested by anyone.

AloisMahdal:
> Keep in mind it's manual script only - it will not be executed automatically

I don't see how this is relevant.


> should not take so much time 

"should not"?  or more like "I wish it would not"? :-)

Looking at the spec, there's actually lot of test cases and some of them would not be easy to set up.

pirat89:
So let's make it simple. I am closing the PR as it doesn't look like it will be merged during next 2 working days. People will have to deal with the problems on their own. If they will not, we can point them that they can download the script manually from github.

PR #77, opened 2018-04-24
Remove the obsoleted system/uefi module

pirat89:
After some testing it was discovered that the module is obsoleted and UEFI seems safe for the update. Messages about GRUB2 are redundant as these are part of the system/grub module. As I don't know details, we need here careful testing to be sure whether we can really remove the module or whether there is some risk about the solution.

The goal of this PR is to have modules for careful testing. In case there will not be found any issue, it can be merged. Review is not required for now.


AloisMahdal:
Some notes from quick meeting with @bocekm: the background is that we used to migrate GRUB1>2 automatically, which failed terribly for UEFI installations.  For that reason, this module existed to simply prevent migration of UEFI systems completely by setting extreme risk.

However, the automated GRUB migration has been since abandoned; nowadays we just leave GRUB1 in place and advise user to migrate it manually. Therefore there is no need for this module to exist.

AloisMahdal:
`qa_ack`: At least, re-run test suite on UEFI system.  Better: talk with UEFI QE expert team and see if there's additional UEFI-specific testing that they could do (manually or by writing automated test).

bocekm:
@pkotvan has performed a manual test and took care to check if:

- there are no changes in /boot/efi except grub config
- UEFI boot order is preserved
- when executing yum update after upgrade, no grub updates are offered

No issues found.

bocekm (dismissed):

pirat89:

bocekm (dismissed):

pirat89:
@AloisMahdal: Lojzo, do you want to run test-suite for our tests here or can I assume it's verified and merge that?

AloisMahdal:
@pirat89 As I said above; re-running on UEFI system is required.

Sorry it takes so long, though.  I've been  burdened by PA suite porting.  Hopefully today I'll be able to finish this one.

AloisMahdal:
Sorry that this takes so long.  We have problems with internal Beaker instance so most jobs take hours to even start and there is higher chance of them aborting; it's hard to make any reasonable progress now.

Also those jobs that we were able to run failed with some strange, seemingly selinux-related issue, we're running more scenarios to be able to confirm whether it's caused by this fix.

AloisMahdal:
Clearing the `qa_ack` granted for previous milestone.

tcerna:
Hi, I tried to test PR77 in release 6.10->7.6 on several machines, but it does not work for me. I will see next week when I'm having meeting with guy from RTT, maybe we will find the problem.

pholica:
Hello guys. I could see some test results done by @tcerna and it seems, that grub package (providing /boot/efi/EFI/redhat/grub.efi) has been removed during the upgrade process. Because of that, the firmware was not able to find the bootloader and boot the system.

For the bootorder and internal Beaker instance, the currently booted boot enty should be always honored when performing reboot. It could be done similar way as it was done in initscripts: https://src.fedoraproject.org/rpms/policycoreutils/pull-request/5
From the test results I could see, it's however not clear if the current boot entry is honored.

pirat89:
Removing from the current milestone and closing th PR as it can be seen that UEFI is not working without additional actions.

tcerna:
Minimum for upgrade on uefi machines:

- Keep files `/boot/efi/EFI/redhat/grub.efi` and `/boot/efi/EFI/redhat/grub.conf`. These files should be on system during upgrade, if they are not there, user can boot to RHEL-7. It is not necessary to have package grub installed, but these files should be stored.
  - This needs support in GRUB module for UEFI machines. There are no corresponding script in `/root/preupgrade/preupgrade-scripts` and no valid script in `/root/preupgrade/postupgrade.d/grub/postupgrade-grub.sh` for this use case.

- There can be also problem with new kernel, because if you save a `/boot/efi/EFI/redhat/grub.conf`, there is information about kernel inside this document.
  - If grubby will be installed and boot-efi will be mounted, it can work.
  - If grub is not installed, it can't work.

- Also it is necessary to implement next boot because of customers and beaker (see comment from pholica). 


How to make sure, the system is uefi and get information for booting:
```
# efibootmgr -v
Boot000C* Red Hat Enterprise Linux 6	HD(1,800,64000,c03123f4-623b-4117-954a-39c71f0e46a4)File(\EFI\redhat\grub.efi)
```

PR #78, opened 2018-04-25
 semanage_export fix syntax

### LINKED ISSUE #39: RestoreConfig: post-upgrade fails
The post-upgrade script uses as input for the semanage commane lines from the file `semanage_export`. But the error occurs during post-upgrade phase when the file contains line as below:

```
fcontext -a -f 'directory' -t root_t '/opt/rh/xyz'
```

The error:
```
semanage fcontext: error: argument -f/--ftype: invalid choice: "'directory'" (choose from 'a', 'f', 'd', 'c', 'b', 's', 'l', 'p')
usage: semanage fcontext [-h] [-n] [-N] [-S STORE] [ --add ( -t TYPE -f FTYPE -r RANGE -s SEUSER | -e EQUAL ) FILE_SPEC ) | --delete ( -t TYPE -f FTYPE | -e EQUAL ) FILE_SPEC ) | --deleteall  | --extract  | --list -C | --modify ( -t TYPE -f FTYPE -r RANGE -s SEUSER | -e EQUAL ) FILE_SPEC ) ]
Could not import 'fcontext -a -f 'directory' -t root_t '/opt/rh/xyz''
```

Similar errors can be seen for other lines, like:
```
Could not import 'boolean -1 xen_use_nfs'
```

jmazanek:
- updating syntax of semanage_export file to
   conform to the RHEL 7 semanage syntax
   

With this fix , postupgrade script restoreConfig.sh is updated, so it updates the command line arguments to semanage in the semanage_export file to RHEL 7 syntax. 
There is no comprehensive list of deprecated options I'm aware of ( man pages are not 100% reliable in this case), so at the moment, only known issues are handled.

Resolves #39

### DIFF
--- a/RHEL6_7/selinux/RestoreConfig/postupgrade.d/restoreConfig.sh
+++ b/RHEL6_7/selinux/RestoreConfig/postupgrade.d/restoreConfig.sh
@@ -15,7 +15,14 @@ semanage_syntax_update() {
 
     for ftype_opt in "${ftype_opts[@]}"
     do
-        sed -i -r 's/'"$ftype_opt"'\s+directory/'"$ftype_opt"' d/g' "$file"
+        sed -i -r "s/$ftype_opt\s+(\x27|\x22)?directory(\x27|\x22)?/$ftype_opt d/g" "$file"
+        sed -i -r "s/$ftype_opt\s+(\x27|\x22)all\s+files(\x27|\x22)/$ftype_opt a/g" "$file"
+        sed -i -r "s/$ftype_opt\s+(\x27|\x22)character\s+device(\x27|\x22)/$ftype_opt c/g" "$file"
+        sed -i -r "s/$ftype_opt\s+(\x27|\x22)block\s+device(\x27|\x22)/$ftype_opt b/g" "$file"
+        sed -i -r "s/$ftype_opt\s+(\x27|\x22)regular\s+file(\x27|\x22)/$ftype_opt f/g" "$file"
+        sed -i -r "s/$ftype_opt\s+(\x27|\x22)?socket(\x27|\x22)?/$ftype_opt s/g" "$file"
+        sed -i -r "s/$ftype_opt\s+(\x27|\x22)symbolic\s+link(\x27|\x22)/$ftype_opt l/g" "$file"
+        sed -i -r "s/$ftype_opt\s+(\x27|\x22)named\s+pipe(\x27|\x22)/$ftype_opt p/g" "$file"
     done
 }
 
--- a/RHEL6_7/selinux/RestoreConfig/postupgrade.d/restoreConfig.sh
+++ b/RHEL6_7/selinux/RestoreConfig/postupgrade.d/restoreConfig.sh
@@ -2,7 +2,26 @@
 
 SEMANAGE_EXPORT_FILE=semanage_export
 
+semanage_syntax_update() {
+
+    ftype_opts=( '-f' '--ftype')
+    bool_opts=( '-1' '-0' '--on' '--off' )
+    local file=$1
+
+    for opt in "${bool_opts[@]}"
+    do
+        sed -i -r 's/'"$opt"'/-m '"$opt"'/g' "$file"
+    done
+
+    for ftype_opt in "${ftype_opts[@]}"
+    do
+        sed -i -r 's/'"$ftype_opt"'\s+directory/'"$ftype_opt"' d/g' "$file"
+    done
+}
+
+
 if test -s "$SEMANAGE_EXPORT_FILE"; then
+    semanage_syntax_update "$SEMANAGE_EXPORT_FILE"
     while read -r line
     do
         # restore saved configuration line by line


jmazanek:
releated issue https://github.com/upgrades-migrations/preupgrade-assistant-modules/issues/39

RHEL6_7/selinux/RestoreConfig/postupgrade.d/restoreConfig.sh
```diff
@@ -2,7 +2,27 @@
...
+
+    for opt in ${ftype_opts[@]}
+    do
+        sed -i -r 's/'"$ftype_opt"'\sdirectory/'"$ftype_opt"' d/g' "$1"
+    done
+}
+
+
 if test -s "$SEMANAGE_EXPORT_FILE"; then
+    if grep -q "release 7\..*" /etc/redhat-release;then
```
pirat89: This is reduntant condition as the script is processed always after the upgrade. It doesn't make sense to check it.

pirat89 (dismissed):

jmazanek:
After testing, the fix seems to work

AloisMahdal:
`qa_ack`: re-run downstream test suite on *all archs* (this was affecting two tests on ppc64 and s390x, IIRC).

pirat89 (dismissed):
Seems good to me.

pirat89:

pirat89 (approved):

RHEL6_7/selinux/RestoreConfig/postupgrade.d/restoreConfig.sh
```diff
@@ -2,7 +2,25 @@
...
+    bool_opts=( '-1' '-0' '--on' '--off' )
+
+    for opt in ${bool_opts[@]}
+    do
+        sed -i -r 's/'"$opt"'/-m '"$opt"'/g' "$1"
+    done
+
+    for opt in ${ftype_opts[@]}
+    do
+        sed -i -r 's/'"$ftype_opt"'\sdirectory/'"$ftype_opt"' d/g' "$1"
```
pirat89: `$ftype_opt` doesn't exist. Probably you meant to use `$opt` or change name on the 15 line.

AloisMahdal:
*preupgrade-assistant-el6toel7-0.7.3-0.201805031451Z.PR78.el6_8.noarch* -- the message is still there.

@pirat89 has already found some issues in the patch; he'll revoke review in few minutes.

RHEL6_7/selinux/RestoreConfig/postupgrade.d/restoreConfig.sh
```diff
@@ -2,7 +2,25 @@
 
 SEMANAGE_EXPORT_FILE=semanage_export
 
+semanage_syntax_update() {
+
+    ftype_opts=( '-f' '-ftype')
```
pirat89: It should be `--ftype`.

AloisMahdal:
More notes on testing: since we have trouble with provisioning, run at least selinux-related tests (notably /CoreOS/vsftpd/Sanity/smoke-write-access) on ppc64.

Note that for PR testing with ipu schedulers, you'll need to host and provide alternative yum repo because COPR's don't support ppc64 and its .repo files contain `$arch` in baseurl, preventing the rpms to be downloaded from there. Ie. ipu tool and test infra can't use PR builds directly.

pirat89 (dismissed):
The fix contains errors. Not sure now whether I missed them before when I did the review or you fixed that and I forgot to update my local branch from the server before I did the rebase (in case you havn't rebased/pulled your local branch yet, you would get that info) , doesn't matter. Please fix it and check it.

@AloisMahdal: apologize for wrong final review, at least after the rebase.

RHEL6_7/selinux/RestoreConfig/postupgrade.d/restoreConfig.sh
```diff
@@ -2,7 +2,25 @@
 
 SEMANAGE_EXPORT_FILE=semanage_export
 
+semanage_syntax_update() {
+
```
AloisMahdal, line 6: I'd also recommend using `local` to prevent defining globals.  (As a side effect, this will allow you to use shorter names.)

AloisMahdal: Another hint: to increase readability and prevent issues caused by later changes, it's good practice to load arguments into variables first and then avoid using positional args.  Ie. don't use `$1` in code, instead do

    local file=$1

and later use `$file`.

RHEL6_7/selinux/RestoreConfig/postupgrade.d/restoreConfig.sh
```diff
@@ -2,7 +2,25 @@
...
+    ftype_opts=( '-f' '-ftype')
+    bool_opts=( '-1' '-0' '--on' '--off' )
+
+    for opt in ${bool_opts[@]}
+    do
+        sed -i -r 's/'"$opt"'/-m '"$opt"'/g' "$1"
+    done
+
+    for opt in ${ftype_opts[@]}
```
AloisMahdal: This should be `"${ftype_opts[@]}"`.  Somebody might add space (or other thing) in the array and your code would break silently.  (Same holds for `bool_opts`.)

RHEL6_7/selinux/RestoreConfig/postupgrade.d/restoreConfig.sh
AloisMahdal: Why `\s`? Is the space guarranteed to be single byte? If so, isn't also type of space guarranteed?  Then you could use just that space and avoid `-r` and also the quoting dance. Just sayin... :)

RHEL6_7/selinux/RestoreConfig/postupgrade.d/restoreConfig.sh
pirat89: @AloisMahdal: I prefer to not use the `file` name in the bash as the file corresponds to the `/usr/bin/file` utility. However I don't have anything against renaming.

jmazanek:
I have probably overlooked those when copying code from my test machine into code repo, which introduced these typos. It is fixed now. Thanks for spotting them

jmazanek:
Notices @AloisMahdal comments just now. Edited. Thanks for pointing out regex error

pirat89 (dismissed):
Now it seems good. There are still mising *local*s but this is on many places of bash scripts, so doesn't matter now.

RHEL6_7/selinux/RestoreConfig/postupgrade.d/restoreConfig.sh
AloisMahdal: I meant using `local` in general, ie. it wouls not just apply particularly to this `file` variable but also `ftype_opts`, `bool_opts`, `opt`, etc.  Anyway, no need to change it here, myfeedback was meant as style guidance.  There would be actually **many** places where this should be changed.

@pirat89 No, one should use concise and short names.  /usr/bin/file is irrelevant here.  Naming *variable* same as *some completely irrelevant file* is not a conflict.

AloisMahdal:
Sorry, still failing.  semanage_export file looks something like this:

    boolean -D
    login -D
    user -D
    port -D
    interface -D
    node -D
    fcontext -D
    fcontext -a -f 'all files' -t public_content_rw_t '/var/ftp/pub/upload.*'

and we get:

    [ 5905.395786] upgrade-post[25594]: semanage fcontext: error: argument -f/--ftype: invalid choice: "'all" (choose from 'a', 'f', 'd', 'c', 'b', 's', 'l', 'p') 
    [ 5905.407089] upgrade-post[25594]: usage: semanage fcontext [-h] [-n] [-N] [-S STORE] [ --add ( -t TYPE -f FTYPE -r RANGE -s SEUSER | -e EQUAL ) FILE_SPEC ) | --delete ( -t TYPE -f FTYPE | -e EQUAL ) FILE_SPEC ) | --deleteall  | --extract  | --list -C | --modify ( -t TYPE -f FTYPE -r RANGE -s SEUSER | -e EQUAL ) FILE_SPEC ) ]

AloisMahdal:
Note that to test this downstream with the same test as I mentioned in last comment:

    ipu -V          # 0.1.11 is latest
    ipu killcache
    ipu upg pam=P78 rut=B /CoreOS/vsftpd/Sanity/smoke-write-access | sh

although this will only work for x86_64, since Copr won't give out .repo files that would be usable for ppc64/s390x. In the latest case, though, this *was* on x86_64.

jmazanek:
@AloisMahdal  I have created a new build with a fix and run the downstream tests. Some of the tests fail, but unless I interpret it incorrectly, those are not related to this PR# and are rather question of the test environment settings. The smoke-write-access , which I assume is a main test does pass now (on x86_64 ). Can you please take a look? Thanks

AloisMahdal:
@jmazanek Unfortunately the test is acutally invalid, because you scheduled it with old ipu version (0.1.9) which schedules 6.9-7.5 by default -- so the internal compatibility checks aborted the whole upgrade.  (The vsftpd passing is bogus in this case; the architecture is "setup case - transform machine - check case" so setup and check both passing is not useful when there was no upgrade.)

At any rate, this has to be tested also on other archs, which due to copr-related limitations needs some ugly hacks so let me take care of that.

AloisMahdal:
OK, so finally, after all the issues, *preupgrade-assistant-el6toel7-0.7.4-0.201806071024Z.PR78.el6_8.noarch* managed to pass on all archs.

Sorry for bad response times lately!

pirat89 (approved):
lgtm

PR #80, opened 2018-05-03
SystemVersion - update versions for rhel-6.10

pirat89:
Start upgrade path from RHEL 6.10.

### DIFF
--- a/RHEL6_7/system/SystemVersion/versions
+++ b/RHEL6_7/system/SystemVersion/versions
@@ -1,8 +1,8 @@
-#release: 6.9
+#release: 6.10
 
 # format: pkgname = N-V-R
 # epoch is not supported now!
 # glibc = glibc-2.17-1.132.el6
 
-glibc = glibc-2.12-1.208.el6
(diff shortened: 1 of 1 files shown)


AloisMahdal:
`qa_ack`: two steps; do both on RHEL-6.9 and RHEL-6.10:

 1. manually run downstream check `check-modules` (a static test that just reads the the versions from file and compares with distro): the check should fail on RHEL-6.9 and pass on RHEL-6.10.

 2. run preupg assessment: the respective module should fail on RHEL-6.9 and pass on RHEL-6.10

AloisMahdal:
Tested with *preupgrade-assistant-el6toel7-0.7.3-0.201805030842Z.PR80.el6_8.noarch*:

    preupg.risk.EXTREME: This is not the latest Red Hat Enterprise Linux 6.10 release.

on RHEL-6.9, and no risk on RHEL=6.10.

The `check-modules` test passed as well.

AloisMahdal (approved):
Not much to say; the change is trivial and tests pass, confirming that the numbers are correct.

Issue #76, opened 2018-04-20
java: exchange of i386 <-> i686


pirat89:
Different names & paths for 32b rpm. Detail description here: https://bugzilla.redhat.com/show_bug.cgi?id=1563407

pirat89:
Important is here function `get_jvm_dir`: https://github.com/upgrades-migrations/preupgrade-assistant-modules/blob/el6toel7/RHEL6_7/system/java/check#L58

~Ugly solution is to include condition to rename `i686` to `i386` always. it's ugly but I don't know how to resolve it reasonable in another way.~ Even this looks like partial fix. We will need to check both. i386 & i686 just to be sure that in future we will not be affected by possible fix of the name.

PR #85, opened 2018-05-30
system/java: handle i686/i386 rpm

pirat89:
In case the java for i686 arch is installed, we need to handle issue when files inside use arch "i386" instead of "i686". It's a black magic. Just use arch-name "i386" when expected jvm dir with "i686" doesn't exist.

Resolves: #76

Testing is not required here for merge as systems should not be affected by this change unless the weird "i686" java is installed on 64b system (which is really weird on RHEL systems).


jmazanek (approved):
lgtm

Issue #84, opened 2018-05-29
system/initscripts/ifcfg: incorrect regexp?


pirat89:
The line below contains extended regular expression which is parsed as basic reular expression. Unless the `+` char should be evaluated as the `+` char, use the `-E` option or fix the regex.

https://github.com/upgrades-migrations/preupgrade-assistant-modules/blob/03d3166fe234c86528bcabca3f52b74388dad668/RHEL6_7/system/initscripts/ifcfg/preupg_network.sh#L36

jmazanek:
https://github.com/upgrades-migrations/preupgrade-assistant-modules/pull/86
fixes the issue

PR #86, opened 2018-05-30
 ifcfg/preupg_network.sh

jmazanek:
- using extended regex


jmazanek:
related issue https://github.com/upgrades-migrations/preupgrade-assistant-modules/issues/84

pirat89 (approved):
LGTM.

Issue #87, opened 2018-05-31
postupgrade.sh looks for grub.conf on ppc64


AloisMahdal:
With preupgrade-assistant-el6toel7-0.7.3-0.201805031451Z.PR78.el6_8.noarch, this can be seen when upgrading on ppc64:

    ...
    [ 9228.421009] upgrade-post[7660]: Running /root/preupgrade/postupgrade.d/grubby_workaround/postupgrade.sh... 
    [ 9228.463151] upgrade-post[7660]: ./postupgrade.sh: line 42: /boot/grub/grub.conf: No such file or directory 
    ...

It should not be because grub.conf is not supposed to be consulted at all on ppc64.

PR #88, opened 2018-05-31
system/grubby - exit not applicable on ppc64 architecture

pirat89:
On the ppc64 architecture on RHEL 6 system is allowed only the Yaboot bootloader which is replaced during the upgrade by the GRUB2 bootloader.

Relevant to RHBZ#1278675 (see the #4 comment):
  https://bugzilla.redhat.com/show_bug.cgi?id=1278675
Fixes: #87


AloisMahdal (approved):
LGTM

Issue #40, opened 2017-11-11
system-requirements depends on dmidecode


AloisMahdal:
Module system-requirements calls dmidecode, which is not available on ppc and s390 architectures.  Currently this results in stray stderr, and probably also module not functioning properly.

PR #89, opened 2018-06-01
system/requirements: check dmidecode and dmesg before use

pirat89:
Some architectures doesn't contain dmidecode utility at all and it does not have to be installed. So to suppress the error msg just check the utility exists. In addition I check even existence of the *dmesg* utility, however it should be installed always on the system as part of the util-linux-ng (on RHEL 6) which is required by the PA/PAM. ~From this point, there is possibly the dead code as the info from the /proc/meminfo file will be never used, but keeping that just to have seatbelts in case we would not get info from the dmesg.~

Additionaly one warning message has been removed, as it was irrelevant for the user (in 99.9% cases will not give any useful info for the user) and the error message has been lowered to warning as it was not an error (here it could gives useful info in some cases, but usually you will not see the message at all).

Fixes: #40

### DIFF
--- a/RHEL6_7/system/requirements/check
+++ b/RHEL6_7/system/requirements/check
@@ -104,26 +104,30 @@ get_free_size() {
 
 get_memory_size() {
   # MEMSIZE is in MB
-  MEMSIZE=$(dmidecode -t 17 \
-    | awk '( /Size/ && $2 ~ /^[0-9]+.B$/ ) {match($2, /^([0-9]+)(.B)$/, arr); $2=arr[1]; $3=arr[2]}
-           ( /Size/ && $3 ~ /^[kK]B$/ ) { unit=1.0/1024 }
-           ( /Size/ && $3 ~ /^MB$/ ) { unit=1 }
-           ( /Size/ && $3 ~ /^GB$/ ) { unit=2**10 }
-           ( /Size/ && $3 ~ /^TB$/ ) { unit=2**20 }
-           ( /Size/ && $2 ~ /^[0-9]+$/ ) { x+=int($2*unit) }
-           END{ print x}')
-  [ -n "$MEMSIZE" ] && return 0
-
-  #maybe virtual machine, so we can get only aroximation of real RAM size
-  log_warning "Can't get memory size with 'dmidecode'! Try 'dmesg' with 1% tolerance."
-  oo=$(dmesg | grep -ioE "System RAM:\s[0-9]+MB")
-  [ $? -eq 0 ] && {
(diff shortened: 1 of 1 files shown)


AloisMahdal:
`qa_ack`: re-run at least some upgrade tests on s390x and ppc64.  There used to be always stderr line in report (/CoreOS/preupgrade-assistant-el6toel7/smoke/upgrade will catch it for sure).

RHEL6_7/system/requirements/check
```diff
@@ -104,26 +104,32 @@ get_free_size() {
...
+             ( /Size/ && $3 ~ /^TB$/ ) { unit=2**20 }
+             ( /Size/ && $2 ~ /^[0-9]+$/ ) { x+=int($2*unit) }
+             END{ print x}')
+    [ -n "$MEMSIZE" ] && return 0
+  }
+
+  # this will not give us exact result, but still ok (~ 1% tolerance usually)
+
+  which dmesg >/dev/null 2>&1 && {
+    oo=$(dmesg | grep -ioE "System RAM:\s[0-9]+MB")
```
jmazanek: On my system (RHEL 6.9, dmesg is installed) this variable is empty (regex does not match). In which case does it match. Also, try to use the variables with descriptive names.

jmazanek (changes requested):

RHEL6_7/system/requirements/check
pirat89: Yeah, that's because the lines are already out of the output the dmesg is providing (it!s still rolling during the run of the system). So you should be able to see that output after some time the system is running but you can loose it and that's why the "third" method to get the data is used (but the last one doesn't provide info about the real installed RAM so well...).

I am keeping the original names of variables as the module shouhld be rewritten. Until that time, I am keeping what is there and fixing only what needs to be really fixed by the PR.

pirat89: Output from the RHEL6 system.
```
# dmesg | grep -ioE "System RAM:\s[0-9]+MB"
System RAM: 2047MB
```

jmazanek (dismissed):

bocekm (approved):
lgtm

AloisMahdal:
OK, so testing with *preupgrade-assistant-el6toel7-0.7.4-0.201806041927Z.PR89.el6_8.noarch*:

 *  [x] static test
 *  [x] other module tests
 *  [x] system-requirement test
 *  [x] smoke upgrade test

pirat89:
Great work! thanks :)

PR #90, opened 2018-06-01
Bump to version 0.7.4

pirat89:
- set upgrade path RHEL 6.10 -> RHEL 7.6
- rename correctly netowrk interfaces regardless amount of existing
  interfaces
- initscripts/ifcfg: set the -E option for grep to process EREG
  correctly (#84)
- system/java: handle corerectly i686 RPMs (#76)
- system/grubby: set as not applicable on ppc64 arch


jmazanek (approved):
lgtm

PR #92, opened 2018-06-05
 system/initscripts/ifcfg - edit udev rules,style enhancement

### LINKED ISSUE #72: system/initscripts/ifcfg: modify module to keep best practices
Because of time, we planned these changes for next milestone:

- [x] replace extreme risks by high risks and create/modify a pre-upgrade script
  to inhibit upgrade in case user haven't fixed that before run of r-u-t
- [x] not sure about the high risk msg when multiple network devices using ethX
  naming are detected - maybe should say in summary what is expected from user
  so user will know that it is expected to check the solution whether want to
  use netX naming or something else.
- [x] use PA API variables inside the preupg_network.sh script

Related: #68

jmazanek:
- the code generating the udev rules was adjusted
   to remove the "kernel" parameter in order
   to disable the unwanted renaming in the
   initramdisk phase.
- style enhancement 
     - the log_extreme_risk is replaced by
       combination of log_high_risk and
       pre-upgrade script, that inhibits
       upgrade.
     - the script preupg_network.sh has
       been edited to use the variables
       provided by PA API when reasonable.
     - Solution text has been edited to
       inform the user about custom naming
       option.

Resolves #72

### DIFF
--- a/RHEL6_7/system/initscripts/ifcfg/check
+++ b/RHEL6_7/system/initscripts/ifcfg/check
@@ -140,10 +140,10 @@ I case 3, to be completely correct you could check that user is not using rhel7 
                 log_slight_risk(full_path + " variable DEVICE is similar to udev predictable network naming scheme. It might cause conflicts.")
                 warning = True
 
-        if ethx_with_addr_count > 1 or ethx_without_addr_count > 1:
+        if ( ethx_with_addr_count  + ethx_without_addr_count) > 1:
             ifnames_to_fix +=1
             warning = True
-        elif ethx_with_addr_count == 1:
+        elif( ethx_with_addr_count + ethx_without_addr_count) == 1:
             log_slight_risk("You use one network device with an old style 'ethX' name.")
             warning = True
 
--- a/RHEL6_7/system/initscripts/ifcfg/check
+++ b/RHEL6_7/system/initscripts/ifcfg/check
@@ -106,9 +106,9 @@ I case 3, to be completely correct you could check that user is not using rhel7 
         script_result.communicate()
         exit_code = script_result.returncode
         if exit_code == 1:
-            log_extreme_risk("The network interface names on the source system are in an inconsistent state, so the upgrade is not possible")
+            log_high_risk("The network interface names on the source system are in an inconsistent state, so the upgrade is not possible")
         elif exit_code == 2:
-            log_extreme_risk("Comment out the MACADDR directive in the ifcfg files and run the Preupgrade Assistant again to enable the upgrade.")
+            log_high_risk("Comment out the MACADDR directive in the ifcfg files and run the Preupgrade Assistant again to enable the upgrade.")
 
     ethx_with_addr_count = 0
     ethx_without_addr_count = 0
--- a/RHEL6_7/system/initscripts/ifcfg/preupg_network.sh
+++ b/RHEL6_7/system/initscripts/ifcfg/preupg_network.sh
@@ -76,7 +76,7 @@ do
         fi
     fi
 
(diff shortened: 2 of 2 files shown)


bocekm (dismissed):

jmazanek:
Made a new build to reflect the issue https://github.com/upgrades-migrations/preupgrade-assistant-modules/issues/72. Summary edited

bocekm (dismissed):

jmazanek:
As discussed off-line with @tcerna , only the upgrade inhibition needs to be tested, as the rest is either tested by the customer or is sufficiently verified by the code review.

How to test:

1. Install the RHEL 6 system with at least 2 network interfaces with eth* kernel naming.
2.  In one of the relevant ifcfg-eth* config files add the "MACADDR" directive which should trigger inhibit.
3. Run the preupgrade-assistant analysis - it should end with needs_action result
4. Run the redhat-upgrade-tool - the upgrade process should be stopped by the pre-upgrade script generated by this fix.

jmazanek:
@tcerna , thanks to your testing, I have spotted additional bug, that wasn't previously caught. The logic would be incorrect if there were exactly 1 ifcfg-ethX with HWADDR and exactly 1 ifcfg-ethX without HWADDR. It probably escaped my attention, because I tested everything on systems with 2+ ifaces.
Can you please test now with a new build and in addition to the steps described in the comment above, also check that the message "preupg.risk.HIGH: You use multiple network devices with old style 'ethX' names." is printed every time there are multiple ifcfg files with DEVICE value and file extension matching the interface name? ( regardless of  number of HWADDR directives)
@bocekm can you please review this change?

bocekm (approved):
The additional fix makes sense.

tcerna:
**Verified in:
  preupgrade-assistant-2.6.0-1.el6.noarch
  preupgrade-assistant-el6toel7-0.7.4-0.201806111315Z.PR92.el6_8.noarch**

1] Have machine with two network interfaces, both have set HWADDR directive, no directive MACADDR is set

* run preupg and see result of 'Ethernet Interface naming':
```
    preupg.risk.SLIGHT: You use one network device with an old style 'ethX' name.
    preupg.log.INFO: On the target system, the eth0 interface will become net0.
    preupg.log.INFO: On the target system, the eth1 interface will become net1.
    preupg.risk.HIGH: You use multiple network devices with old style 'ethX' names.
```

* run redhat-upgrade-tool, it ended correctly with possibility of reboot and updade to rhel-7.5:
```
    Shutting down interface eth0:  [  OK  ]
    Shutting down loopback interface:  [  OK  ]
    Bringing up loopback interface:  [  OK  ]
    Bringing up interface net0:  
    Determining IP information for net0... done.
    [  OK  ]
    Finished. Reboot to start upgrade.
```

2] Have machine with two network interfaces, both have set HWADDR and MACADDR directives

* run preupg and see result of 'Ethernet Interface naming':
```
    preupg.risk.SLIGHT: You use one network device with an old style 'ethX' name.
    preupg.risk.HIGH: Comment out the MACADDR directive in the ifcfg files and run the Preupgrade Assistant again to enable the upgrade.
    preupg.risk.HIGH: You use multiple network devices with old style 'ethX' names.
```

* run redhat-upgrade-tool, it ended correctly with exit code 1. it stopped on script invalid_config_check because of directive MACADDR:
```
Invalid network configuration detected, stopping upgrade
Following preupgrade script(s) failed:
/root/preupgrade/preupgrade-scripts/invalid_config_check.sh exited with status 1
exiting
```

3] Verified also output of this module from preupg tool. There were two ethernet interfaces where one has set only MACADDR directive and second only HWADDR directive

* run preupg and see result of 'Ethernet Interface naming':
```
    preupg.risk.SLIGHT: You use one network device with an old style 'ethX' name.
    preupg.risk.SLIGHT: /etc/sysconfig/network-scripts/ifcfg-eth1 is old style ethX name without HWADDR, its name can change after the upgrade.
    preupg.risk.HIGH: Comment out the MACADDR directive in the ifcfg files and run the Preupgrade Assistant again to enable the upgrade.
    preupg.risk.HIGH: You use multiple network devices with old style 'ethX' names.
```

PR #96, opened 2018-07-01
 system/pam

jmazanek:
- fixed the issue from BZ# 1463710
 - with this fix module no longer
   hangs and correctly comments
   out deprecated directives in
   "/etc/pam.d/*" configuration files
 - rewritten for better readability

### DIFF
--- a/RHEL6_7/system/pam/check
+++ b/RHEL6_7/system/pam/check
@@ -4,22 +4,57 @@
 
 #END GENERATED SECTION
 
-POSTUPGRADE_DIR="$VALUE_TMP_PREUPGRADE/postupgrade.d/pam"
+POSTUPGRADE_DIR="$POSTUPGRADE_DIR/pam"
+CLEANCONF_DIR="$VALUE_TMP_PREUPGRADE/cleanconf/etc/pam.d"
+
 if [[ ! -d "$POSTUPGRADE_DIR" ]]; then
     mkdir -p "$POSTUPGRADE_DIR"
 fi
+if [[ ! -d "$CLEANCONF_DIR" ]]; then
+    mkdir -p "$CLEANCONF_DIR"
+fi
+
 SCRIPT_NAME="postupgrade-pam.sh"
 POST_SCRIPT="postupgrade.d/$SCRIPT_NAME"
+
 cp -f $POST_SCRIPT $POSTUPGRADE_DIR/$SCRIPT_NAME
+
 fail=0
 
+comment_invalid () {
+    perl -p -i -e 's/^/#/g if /^(?!.*#)(?=.*(pam_passwdqc|pam_ecryptfs))/;' "$1"
+}
+check_deprec_dirctv () {
+
+    deprec_dirctv="$1"
+    input_file="$2"
+    export deprec_dirctv
+    export input_file
+    perl -ne '$deprec_dirctv = $ENV{deprec_dirctv};$input_file = $ENV{input_file};
+    print "PAM: The $input_file file contains the $deprec_dirctv module, which is no longer supported.\n"
+    if /^(?!.*#)(?=.*$deprec_dirctv)/' "$input_file"
+
+}
+copy_clean () {
+    cp -p "$1" "$2"
+}
+
 #pam_passwdqc and pam_ecryptfs were removed
-for file in /etc/pam.d/*;
+for file in $(find -P /etc/pam.d/ -maxdepth 1 -type f)
 do
-  grep pam_passwdqc "$file" | grep -q -v ^# && sed -e '/pam_passwdqc/s/^/#/g' >"$VALUE_TMP_PREUPGRADE/cleanconf/$file" && log_medium_risk "PAM: The $file file contains the pam_passwdqc module, which is no longer supported." && fail=1
-  grep pam_ecryptfs "$file" | grep -q -v ^# && sed -e '/pam_ecryptfs/s/^/#/g' >"$VALUE_TMP_PREUPGRADE/cleanconf/$file" && log_high_risk "PAM: The $file file contains the pam_ecryptfs module, which is no longer supported. Your ecryptfs encrypted files will not work after the upgrade. Decrypt the files before the upgrade and possibly switch to a different encryption." && fail=1
+    clean_file="$CLEANCONF_DIR"/$(basename "$file")
+
+    deprec_status=$(check_deprec_dirctv "pam_passwdqc" "$file")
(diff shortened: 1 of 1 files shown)


jmazanek:
How to test:

1. In one or more of the configuration files in /etc/pam.d/ directory, add the one or more deprecated directives (pam_passwdqc|pam_ecryptfs)
example:
`password    required      pam_passwdqc.so`

2. Run the p-a modules with this fix.

3. Results should be `needs_inspection` if only `pam_passwdqc` directive is present, `needs_action` if both or only `pam_ecryptfs` and `pass` otherwise. Configuration files with deprecated directives commented out will be copied into `/root/preupgrade/cleanconf/etc/pam.d/`.

RHEL6_7/system/pam/check
```diff
@@ -4,22 +4,57 @@
...
 
+comment_invalid () {
+    perl -p -i -e 's/^/#/g if /^(?!.*#)(?=.*(pam_passwdqc|pam_ecryptfs))/;' "$1"
+}
+get_deprec_dirctv () {
+
+    deprec_dirctv="$1"
+    file="$2"
+    export deprec_dirctv
+    export file
```
pirat89: why do you export the file variable? By this you are rewritting the `$file` variable outside of the scope of this function (even when in the current call you are providing the outer `$file` content, storing it into the innter `$file` variable). Both of these exports are not needed as the whole perl script can be written without ENV at all.

RHEL6_7/system/pam/check
```diff
@@ -4,22 +4,57 @@
...
+}
+get_deprec_dirctv () {
+
+    deprec_dirctv="$1"
+    file="$2"
+    export deprec_dirctv
+    export file 
+    deprec_status=$(perl -ne '$deprec_dirctv = $ENV{deprec_dirctv};$file = $ENV{file}; 
+    print "PAM: The $file file contains the $deprec_dirctv module, which is no longer supported.\n"
+    if /^(?!.*#)(?=.*$deprec_dirctv)/' "$2")
```
pirat89: You should use everywhere the `$file` variable  or everywhere the `$2` variable. Do not mix them. In addition append indentation for inner code to improve readability. But much rather I would see deprec_status omitted completely for use outside of the scope of the function. You can just examine whether the deprecated directive is use in the given file or not and base on that print the output (that you can store outside of this function). IOW, the `deprec_status` should not be used here, but outside of the function.

RHEL6_7/system/pam/check
```diff
@@ -4,22 +4,57 @@
...
+
 #pam_passwdqc and pam_ecryptfs were removed
-for file in /etc/pam.d/*;
+for file in $(find -P /etc/pam.d/ -maxdepth 1 -type f)
 do
-  grep pam_passwdqc "$file" | grep -q -v ^# && sed -e '/pam_passwdqc/s/^/#/g' >"$VALUE_TMP_PREUPGRADE/cleanconf/$file" && log_medium_risk "PAM: The $file file contains the pam_passwdqc module, which is no longer supported." && fail=1
-  grep pam_ecryptfs "$file" | grep -q -v ^# && sed -e '/pam_ecryptfs/s/^/#/g' >"$VALUE_TMP_PREUPGRADE/cleanconf/$file" && log_high_risk "PAM: The $file file contains the pam_ecryptfs module, which is no longer supported. Your ecryptfs encrypted files will not work after the upgrade. Decrypt the files before the upgrade and possibly switch to a different encryption." && fail=1
+    clean_file="$CLEANCONF_DIR"/$(basename "$file")
+
+    get_deprec_dirctv "pam_passwdqc" "$file"
```
pirat89: Here you would set the `deprec_dirctv` or `deprec_status` variable:
```
  deprec_status=$(get_deprec_dirctv "pam_passwdqc" "$file")
```

RHEL6_7/system/pam/check
```diff
@@ -4,22 +4,57 @@
...
 POST_SCRIPT="postupgrade.d/$SCRIPT_NAME"
+
 cp -f $POST_SCRIPT $POSTUPGRADE_DIR/$SCRIPT_NAME
+
 fail=0
 
+comment_invalid () {
+    perl -p -i -e 's/^/#/g if /^(?!.*#)(?=.*(pam_passwdqc|pam_ecryptfs))/;' "$1"
+}
+get_deprec_dirctv () {
```
pirat89: From the behaviour of the function, you do not return `deprec_dirctv` but `deprec_status`. Much better name would be `detect_deprec_dictv` or `check_deprec_dirctv`.

pirat89 (dismissed):
I suggest to modify the current solution. And remove please trailing whitespaces.

RHEL6_7/system/pam/check
jmazanek: I have renamed the variable so there is no name conflict with another variable used outside the function. Thanks for spotting it! Export  as such is not a problem, and using it + $ENV{var} construct is a standard way to pass shell variable to perl.

RHEL6_7/system/pam/check
jmazanek: Fixed, I'm using renamed "$input" variable on all places. Not mixing anymore

RHEL6_7/system/pam/check
jmazanek: Fixed the variable naming, so this is no longer an issue

RHEL6_7/system/pam/check
jmazanek: renamed function to check_deprec_dirctv

RHEL6_7/system/pam/check
pirat89: This is still issue as this is not related to variable naming.

pirat89 (approved):
lgtm. I have done just code review without testing of the module functionality. But from the description it looks good.

~Just one unknown: what in case there is not found standard file but directory or symlink? Not sure how much is it relvant to have it one of those instead of standard file, but...~ I overloked the type argument in the find command.

jmazanek:
Subdirectory should not, to my knowledge appear there.  This line in the code should ensure, that inputs are just ordinary files `for file in $(find -P /etc/pam.d/ -maxdepth 1 -type f)`

AloisMahdal:
`qa_ack`: test according to description from @jmazanek

tcerna:
New test case was written for this PR96:
preupgrade-assistant-el6toel7/ast/pam-deprecated-directives

Test passed in preupgrade-assistant-el6toel7-0.7.5-0.201807021157Z.PR96.el6_8.noarch
-> VERIFIED

PR #97, opened 2018-07-03
 system/BinariesRebuild

jmazanek:
- added test to checking
   if the variable passed
   to the gen_blacklist
   function is not empty
   before testing if it
   is a file.

### DIFF
--- a/RHEL6_7/system/BinariesRebuild/check
+++ b/RHEL6_7/system/BinariesRebuild/check
@@ -23,6 +23,7 @@ get_blacklist_file() {
 gen_blacklist() {
   local blacklist_cfg=$(get_blacklist_file)
   [ $? -ne 0 ] && return 1
+  [ -z "$blacklist_cfg" ] && return 1
   [ ! -f "$blacklist_cfg" ] && {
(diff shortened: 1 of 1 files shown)


jmazanek:
This fix tests if the argument to gen_blacklist function is not empty, before passing it to test if it is an existing file. The strange quotation format  on the second line is internal to the error message, that seems to be produced by `ldd` command when missing the expected library.

jmazanek:
How to test:
1. In /etc/preupgrade-assistant.conf add a following input:

    [xccdf_preupg_rule_system_BinariesRebuild_check]
    exclude_file=

2. Run the preupgrade-assistant , the warning message for the missing file should not be printed

3. As a control case, put a value to exclude_file directive to some non-existing file. In this case the warning message will be printed.

pirat89 (approved):
LGTM.

tcerna:
New test case for this PR created:
/CoreOS/preupgrade-assistant-el6toel7/ast/binaries-rebuild-checks-exclude-file-directive

Test case passed with installed preupgrade-assistant-el6toel7-0.7.5-0.201807030924Z.PR97.el6_8.noarch
-> VERIFIED

PR #104, opened 2018-09-04
Silence grep test for AuthorizedKeysCommand

AloisMahdal:
Modules are not supposed to generate stdout or stderr on their own.


AloisMahdal:
This is how the module looks like now in the result.html.

ssh-keycat

pirat89:
Thanks!

pirat89 (approved):

PR #105, opened 2018-09-05
fix used modes for os.chmod in python's scripts

pirat89:
In python, use of 0xxx mode is incorrect. Either xxx is wrong as
mode is not set as expected. Only correct way here is use of 0oxxx
format. E.g.:
        0o775


pirat89:
See: https://stackoverflow.com/questions/5231901/permission-problems-when-creating-a-dir-with-os-makedirs-in-python/5231994#5231994

This change should not require additional testing as everything is running under root that is ignoring DAC.

jmazanek (approved):
LGTM

AloisMahdal (approved):

AloisMahdal:
LGTM.  I did not actually know about the `0oXXX` notation, good to know.

PR #110, opened 2018-10-05
Bump to version 0.7.7

pirat89:
- services/tomcat: migrate config files and webapps to the new paths
- networking/samba: improve the informational message


shaded-enmity (approved):
LGTM

PR #117, opened 2019-01-14
Add check for Linux Disk Layout usage on s390x arch

artmello:


RHEL6_7/storage/dasd-format/check
```diff
@@ -0,0 +1,97 @@
...
+{
+	IFS=$STR_IFS
+
+	exit $1
+}
+
+[ "$ARCH" == "s390x" ] || wrap_exit $RESULT_NOT_APPLICABLE
+test -x $DASDVIEW || wrap_exit $RESULT_NOT_APPLICABLE
+test -x $GREP || wrap_exit $RESULT_NOT_APPLICABLE
+test -x $CUT || wrap_exit $RESULT_NOT_APPLICABLE
```
pirat89: You cannot write that result is `$RESULT_NOT_APPLICABLE` when `grep` or `cut` utilities are missing. It seems OK for `dasdview` as when this one is not installed, probably the module is not applicable. But the others are requirements to be able to process it correctly. From that point, you should specify those two in the `module.ini` file as `binary_req` - as when these two are missing, it is supposed to ask user to install them and run scan of the system again. See the [description here](https://github.com/upgrades-migrations/preupgrade-assistant/blob/master/doc/module_writing_tutorial/01_simple_informational_module/module.ini#L38). From that point, you don't need to bother about existence of these two utilities then, so you can use them just calling `grep` and `cut` without need of `$GREP` and `$CUT` variables. But you can keep them, it's not problem.

RHEL6_7/storage/dasd-format/check
```diff
@@ -0,0 +1,97 @@
...
+
+
+function detect_linux_disk_layout()
+{
+	log_debug "Checking Linux Disk Layout for a device $1"
+
+	local format
+	format=$($DASDVIEW -x $1 2>/dev/null | $GREP "format :" | $CUT -f 3)
```
pirat89: Not sure that subcommand is correct. Probably there is used `\t` as a separator but in that case, regexp is incorrect and cut as well (in case regexp would be correct, it would return just `hex` substring). Example of the line (probably tabulators have been replaced by spaces already:
```
format                 : hex 2          dec 2           CDL formatted
```

In such case, the output would be not catched. From the code itself, I guess you want to get directly the `LDL`/`CDL` (`-f 7`) substring or everything after hex (`-f 3-`). Anyway, it seems simpler to me just get the line and then grep it for LDL or CDL substring (comment below).
```suggestion
	format=$($DASDVIEW -x $1 2>/dev/null | $GREP "^[:spaces:]*format[:spaces:]*:")
```

RHEL6_7/storage/dasd-format/check
```diff
@@ -0,0 +1,97 @@
...
+
+	test -n "$format" || {
+		log_error "dasdview parsing failed for a device $1"
+		wrap_exit $RESULT_ERROR
+	}
```
pirat89, line 28: I am not so familiar with DASD, but is it possible that you would have disk unrelated to DASD? In such case you would not recieve probably any `format` output. Maybe completely wrong comment as I loose what I know about DASD soon after I learn something about that (:disappointed:) so in case it's irelevat, just ignore it.

RHEL6_7/storage/dasd-format/check
```diff
@@ -0,0 +1,97 @@
...
+
+	log_debug "Detected Direct Access Storage Device format: $format"
+
+	[[ "$format" == *"LDL"* ]]
```
pirat89: Why the asterisks? That's unkown to me. I guess it should detect substring, but I would afraid it will try to expand path (filename) as from usual behaviour of bash. Anyway, I suggest to use rather just grep:
```suggestion
	echo "$format" | $GREP -q "[:space:]LDL[:space:]"
```

RHEL6_7/storage/dasd-format/check
```diff
@@ -0,0 +1,97 @@
...
+	log_info "Check system devices for Linux Disk Layout use."
+
+	IFS=$'\n'
+
+	for dev in $(list_devices) ; do
+		log_debug "Checking device '$dev'"
+		check_device "$dev"
+		if [ $? -ne 0 ]; then
+			log_error "A Direct Access Storage Device (DASD) was detected using Linux Disk Layout (LDL) format: $dev"
+			log_high_risk "The in-place upgrade cannot proceed further. Linux Disk Layout (LDL) format is not supported on Red Hat Enterprise Linux 7 and in-place upgrade can cause data loss."
```
pirat89: You cannot combine `log_error` and `log_high_risk`. This points more like to something that should be reported than output for user that he is not able to do upgrade. In addition, you want to inhibit the in-place upgrade so you should use `log_extreme_risk ` (the `log_high_risk` function is telling to the user, that they should do (or check) something manually before the upgrade, but can continue). Please, use `log_extreme_risk` in both cases. Maybe just something like: `A Direct Access Storage Device (DASD) was detected using Linux Disk Layout (LDL) format: $dev". Upgrade is not possible.` would be ok as more info is part of the solution.txt file. But you can keep both of them, just use `log_extreme_risk`.

pirat89 (changes requested):
First round is done :) Please, modify the script. In addition, the `function strin()` is kind of obsoleted or discouraged. Not sure why already (maybe it was related to our experiments with older systems) anyway, use for functions shorter format please:
```
funcname()
{
  ...
}
```

pirat89:
Anyway, in case of `cut` - when you are not sure about number of whitespaces, it makes sense to use `awk '{print $<number>}'` instead. It will do the thing for you as well.

artmello:
hey @pirat89, thanks for reviewing. Hopefully all your comments were addressed on latest commit, let me know what do you think.

RHEL6_7/storage/dasd-format/check
```diff
@@ -0,0 +1,88 @@
...
+
+	log_info "Check system devices for Linux Disk Layout use."
+
+	IFS=$'\n'
+
+	for dev in $(list_devices) ; do
+		log_debug "Checking device '$dev'"
+		check_device "$dev"
+		if [ $? -ne 0 ]; then
+			log_high_risk "A Direct Access Storage Device (DASD) was detected using Linux Disk Layout (LDL) format: $dev. Upgrade is not possible."
```
pirat89: Possibly I did mistake in long explanation comment last time here. To be clear, we need to inhibit the upgrade, so `log_extreme_risk` should be used:
```suggestion
			log_extreme_risk "A Direct Access Storage Device (DASD) was detected using Linux Disk Layout (LDL) format: $dev. Upgrade is not possible."
```

RHEL6_7/storage/dasd-format/module.ini
```diff
@@ -0,0 +1,6 @@
+[preupgrade]
+content_title = Detection of Direct Access Storage Device (DASD) format on s390x platform for LDL format
+content_description = On the s390x architecture, if a partition is formatted with LDL on one or more Direct Access Storage Devices, the upgrade process will fail and cause a data loss on such partition.
+author = Arthur Mello <amello@redhat.com>
+requires = grep, s390utils-base, util-linux
+binary_req = blkid, grep
```
pirat89, line 6: It doesn't make sense to require grep rpm and grep binary.  Same for blkid. Require just rpm or binary. I prefer rpm requirement only, but you can choose any of them you prefer.

The problem is in `s390utils-base` requirement. In such case, you would get automatically HIGH risk and `action_needed` result on all non-s390x archs. But you would use `applies_to` for the RPM so module will be automatically applied only when the RPM will be installed (so you will not need the check of `DASDVIEW` inside the script on your own).

pirat89 (changes requested):

artmello:
@pirat89 thanks again for reviewing. Hopefully the comments are addressed with latest commit

pirat89:
Seems OK. In case you not find any new issues, squash commits please :)

pirat89:
--test--

AloisMahdal:
Tested with *preupgrade-assistant-el6toel7-0.7.7-0.201901211333Z.PR117.el6_8.noarch*.  Manually and by running regression test suite. No new issues.

pirat89 (approved):

pirat89:
@AloisMahdal thanks Lojza :)

PR #118, opened 2019-02-12
Bump to version 0.7.8

pirat89:
- Add check for Linux Disk Layout usage on s390x arch


bocekm (approved):

pirat89:
The CI is down. Merging anyway.

PR #119, opened 2019-02-14
Remove legacy grub before installation of grub2

pirat89:
The grub2-common rpm has file conflict with the grub rpm so
installation of grub2 is not possible. This occurs when rhelup
is called without the --cleanup-post option.


bocekm (dismissed):
Tested and it works.

bocekm (changes requested):
The link to the system admin guide in the solution.txt is not working. The correct one is https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/system_administrators_guide/sec-upgrading_from_grub_legacy_to_grub_2

It would actually work also if you only add underscore between grub and 2 ;)

pirat89:
@bocekm Good catch!

bocekm (approved):

Issue #122, opened 2019-06-25
system/tuned: stray output


AloisMahdal:
During the pre-upgrade scan, tuned module prints following (abridged) to stdout:

    `/etc/tune-profiles//active-profile' -&gt; `/root/preupgrade/postupgrade.d/tuned/active-profile'
    `/etc/tune-profiles//default' -&gt; `/root/preupgrade/postupgrade.d/tuned/default'
    `/etc/tune-profiles//default/ktune.sysconfig' -&gt; `/root/preupgrade/postupgrade.d/tuned/default/ktune.sysconfig'
    `/etc/tune-profiles//default/sysctl.ktune' -&gt; `/root/preupgrade/postupgrade.d/tuned/default/sysctl.ktune'
    [...]
    `/etc/tune-profiles//virtual-host' -&gt; `/root/preupgrade/postupgrade.d/tuned/virtual-host'
    `/etc/tune-profiles//virtual-host/ktune.sh' -&gt; `/root/preupgrade/postupgrade.d/tuned/virtual-host/ktune.sh'
    `/etc/tune-profiles//virtual-host/ktune.sysconfig' -&gt; `/root/preupgrade/postupgrade.d/tuned/virtual-host/ktune.sysconfig'
    `/etc/tune-profiles//virtual-host/sysctl.ktune' -&gt; `/root/preupgrade/postupgrade.d/tuned/virtual-host/sysctl.ktune'
    `/etc/tune-profiles//virtual-host/tuned.conf' -&gt; `/root/preupgrade/postupgrade.d/tuned/virtual-host/tuned.conf'

This goes along with medium risk:

    preupg.risk.MEDIUM: Tuned profiles were detected in the /etc/tune-profiles/ directory. See the tuned-adm and tuned-profiles man pages for more information.

Version: *preupgrade-assistant-el6toel7-0.7.9-1.el6*

AloisMahdal:
AKA OAMG-407 in downstream tracker.

PR #125, opened 2020-05-25
system/tuned: updated the check, added more rules

yarda:
Related: https://projects.engineering.redhat.com/browse/OAMG-3185

Signed-off-by: Jaroslav Škarvada <jskarvad@redhat.com>

It adds more check logic. Currently the code doesn't work for profiles with spaces in the name (I am not sure whether it's supported, if yes, feel free to update the code).

It works the following way:
- it backups profiles which it doesn't recognize (i.e. profiles we don't ship)
- it backups profiles that we ship in case they were modified
- profiles that we ship are not backed up if they weren't modified
- backed up profiles needs to be hand migrated
- it maps active_profile to the known replacement profile
- if it cannot recognize the active_profile or if it is not supported, it uses the default recommended profile

Also it wouldn't be bad if it could run check after the migration (i.e. on the RHEL-7). In such case I would recommend running `tuned-adm recommend` (on the RHEL-7) and if its output differs from the $ACTIVE_PROFILE it could warn user that her or his profile is different from the recommended one and that she or he could reconsider using the recommended defaults.

### DIFF
--- a/RHEL6_7/system/tuned/check
+++ b/RHEL6_7/system/tuned/check
@@ -4,6 +4,15 @@
 
 #END GENERATED SECTION
 
+FILES=$(rpm -ql tuned 2> /dev/null) && HAS_TUNED=1
+VERIFY=$(rpm -qV tuned 2> /dev/null)
+FILES="$FILES $(rpm -ql tuned-profiles-sap 2> /dev/null)"
+VERIFY="$VERIFY $(rpm -qV tuned-profiles-sap 2> /dev/null)"
+FILES="$FILES $(rpm -ql tuned-profiles-sap-hana 2> /dev/null)"
+VERIFY="$VERIFY $(rpm -qV tuned-profiles-sap-hana 2> /dev/null)"
+FILES="$FILES $(rpm -ql tuned-profiles-oracle 2> /dev/null)"
+VERIFY="$VERIFY $(rpm -qV tuned-profiles-oracle 2> /dev/null)"
+
 TUNED_DIR="/etc/tune-profiles/"
 if [[ -d "$TUNED_DIR" ]]; then
     POSTUPGRADE_TUNED_DIR="$VALUE_TMP_PREUPGRADE/postupgrade.d/tuned"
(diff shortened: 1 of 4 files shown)


pirat89 (changes requested):
discussed in person, we will do some changes here:
- update solution.txt
- create postupgrade script
- do backup into different dirs

just putting the note here to inform that no other review is needed

Edit: to make it clear, one of original problems and still the current one, that inside the pustupgrade dir (when backup happens), if any executable is stored (from rhel 7 system), it will be processed during the post-upgrade phase. which we do not want in this case.

AloisMahdal:
Note that the stray output is #122.

pirat89:
@yarda kind of done. I did one push force to fix some changes to fix couple of things in those two additional commits. So far it seems it works. Pls check the current solution and let me know.

RHEL6_7/system/tuned/zz_tuned_post.sh
```diff
@@ -0,0 +1,30 @@
...
+    # ok, tuned probably doesn't work correctly right now in the post-upgrade
+    # "mode", let's keep it on user to follow remediation instruction from
+    # the preupgrade report
+    echo >&2 "Info: cannot get the recommended set up for tuned right now. Skip."
+    exit 0
+fi
+
+
+ACTIVE_PROFILE=$(cat "/etc/tuned/active_profile")
+if [ "$ACTIVE_PROFILE" != "$REC_PROFILE" ]; then
```
yarda: If tuned hasn't been started yet then /etc/tuned/active_profile is "" and  /etc/tuned/profile_mode is "auto", in such case the recommended profile will be automatically applied and the test should handle this case.

yarda: I.e. it shouldn't write the warning message if /etc/tuned/active_profile is "" and /etc/tuned/profile_mode is "auto"

RHEL6_7/system/tuned/solution.txt
```diff
@@ -1,4 +1,4 @@
-The module stores the custom 'tuned' profiles, which are located in the /etc/tune-profiles/ directory, to POSTUPGRADE_DIR.
-
-After the upgrade to Red Hat Enterprise Linux 7, the profiles can be migrated manually with the help of the 'tuned-man' and 'tuned-profiles' commands to the /etc/tuned/ directory.
+After the upgrade to Red Hat Enterprise Linux (RHEL) 7, the profiles can be migrated manually with the help of the 'tuned-man' and 'tuned-profiles' commands to the /etc/tuned/ directory.
```
yarda: It's `tuned` and `tuned-profiles` manual pages not commands. Also Tuned documentation (https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html-single/performance_tuning_guide/index#chap-Red_Hat_Enterprise_Linux-Performance_Tuning_Guide-Tuned) could be mentioned.

RHEL6_7/system/tuned/solution.txt
```diff
@@ -1,4 +1,4 @@
...
+It is recommended to execute the "tuned-adm recommend" command after the upgrade on the RHEL 7 system. If the output differs from the active profile, reconsider change of the active profile to the recommended one.
```
yarda: Maybe this recommendation could be dropped, because this check is now automatically run in post.

RHEL6_7/system/tuned/solution.txt
```diff
@@ -1,4 +1,4 @@
...
+Customized and third party tuned profiles are backed up in 'TUNED_BACKUP_DIR'.
```
yarda, line 2: I would drop the old Fedora power management guide and would link the RHEL performance guide mentioned above.

RHEL6_7/system/tuned/check
```diff
@@ -4,20 +4,30 @@
...
+    VERIFY="$VERIFY $(rpm -qV "$pkg" 2> /dev/null)"
+done
 
 TUNED_DIR="/etc/tune-profiles/"
 if [[ -d "$TUNED_DIR" ]]; then
-    POSTUPGRADE_TUNED_DIR="$VALUE_TMP_PREUPGRADE/postupgrade.d/tuned"
-    if [[ ! -d "$POSTUPGRADE_TUNED_DIR" ]]; then
-        mkdir -p "$POSTUPGRADE_TUNED_DIR"
+    TUNED_BACKUP_DIR="$VALUE_TMP_PREUPGRADE/backup/tuned"
+    if [[ ! -d "$TUNED_BACKUP_DIR" ]]; then
```
yarda: You could drop the `if` conditional, because `mkdir -p DIR` always succeed even if the directory exists.

yarda (commented):
In overall LGTM, please address my comments (me as the original PR author cannot approve or request the changes :)

pirat89 (dismissed):
lgtm. Thanks @yarda for your contribution! :)

AloisMahdal:
I ran the original downstream test with

    preupgrade-assistant-2.6.1-1.201903271312Z.master.el6_10
    preupgrade-assistant-el6toel7-0.7.9-0.202007130935Z.PR125.el6_10
    preupgrade-assistant-el6toel7-data-0.20200704-1.el6_10
    redhat-upgrade-tool-0.8.0-1.201903271313Z.rhel6.el6_10

Is this expected in upgrade.log?:

    [..]
    [..]
    Jul 13 12:11:14 (hostname) upgrade-post[31894]: Config file /var/lib/logrotate.status does not exists.
    Jul 13 12:11:14 (hostname) upgrade-post[31894]: Copy file var/lib/logrotate.status to /var/lib/logrotate.status
    Jul 13 12:11:14 (hostname) upgrade-post[31894]: Running /root/preupgrade/postupgrade.d/zz_tuned_post.sh...
    Jul 13 12:11:14 (hostname) upgrade-post[31894]: Warning: The recommended tuned profile 'throughput-performance' is different from the active one 'virtual-guest'. Consider the change of the profile.
    Jul 13 12:11:14 (hostname) upgrade-post[31894]: redhat_upgrade_tool.util WARNING: failed to remove /system-upgrade-root: [Errno 16] Device or resource busy: '/system-upgrade-root/mnt'
    Jul 13 12:11:15 (hostname) upgrade-post[31894]: resetting bootloader config
    [..]
    [..]

Asking just about the "recommended tuned profile"; the "failed to remove" is already known as upgrades-migrations/redhat-upgrade-tool#52

Otherwise no issues found; the test passed.

pirat89:
@AloisMahdal it's expected message. The warning is additional pointer that user should check it. The same instruction is already in the preupgrade report.

AloisMahdal:
also note that this message is not in preupgrade asessment (result.html); the result there is `pass`.

yarda:
Yes, the warning about recommended profile is usually OK, but in this case it's suspicious.

In case it's virtual machine, the 'tuned-adm recommend' should recommend 'virtual-guest' profile, not the 'throughput-performance' profile. What's the output of the 'virt-what' command on the machine?

AloisMahdal:
*virt-what* produces empty output on this one.

yarda:
So it's probably OK, on bare metal (with the empty virt-what output) the recommendation to use 'throughput-performance' profile is correct.

AloisMahdal:
> Also it wouldn't be bad if it could run check after the migration (i.e. on the RHEL-7). In such case I would recommend running
> tuned-adm recommend (on the RHEL-7) and if its output differs from the $ACTIVE_PROFILE it could warn user that her or his > profile is different from the recommended one and that she or he could reconsider using the recommended defaults.

Why this?  IOW, how does having different than recommended profile relate to upgrade?

If it's because recommendations changed, could we predict that change and report the fact to the assessment?

yarda:
Theoretically we cannot predict it. Tuned on RHEL-6 doesn't have the recommend feature. On RHEL-7 the 'recommend' behaviour can be also changed by 3rd party packages.

AloisMahdal:
OK, talking to @pirat89 the intent is that user is already instructed about possible need of re-consideration.  This warning is a smart reminder of that.

However there's actually nothing about this in the assessment:

oamg3352-ast

AloisMahdal:
With

    preupgrade-assistant-2.6.1-1.201903271312Z.master.el6_10
    preupgrade-assistant-el6toel7-0.7.9-0.202007131641Z.PR125.el6_10
    preupgrade-assistant-el6toel7-data-0.20200704-1.el6_10
    redhat-upgrade-tool-0.8.0-1.201903271313Z.rhel6.el6_10

the original downstream test now passed.

pirat89 (approved):

pirat89:
Thanks guys @yarda @AloisMahdal . Merging

PR #128, opened 2020-07-07
system/SysconfigCgroupDaemon: handle dead symlinks

pirat89:
The module crashes on dead symlink under /etc/sysconfig
```python
Traceback (most recent call last):
  File "/root/preupgrade/RHEL6_7/system/SysconfigCgroupDaemon/check", line 72, in <module>
    if cgroup_daemon():
  File "/root/preupgrade/RHEL6_7/system/SysconfigCgroupDaemon/check", line 62, in cgroup_daemon
    with open(root+"/"+filename, "r") as f:
IOError: [Errno 2] No such file or directory: '/etc/sysconfig/deadlink'
```

See the RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1851322
Thanks to @rmetrich for the patch.

### DIFF
--- a/RHEL6_7/system/SysconfigCgroupDaemon/check
+++ b/RHEL6_7/system/SysconfigCgroupDaemon/check
@@ -41,8 +41,14 @@ def cgroup_daemon():
 	found = False
 	for root, dirs, files in os.walk('/etc/sysconfig'):
 		for filename in files:
-			with open(root+"/"+filename, "r") as f:
-				data = f.read()
+			try:
+				with open(root+"/"+filename, "r") as f:
+					data = f.read()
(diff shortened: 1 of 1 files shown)


version
```diff
@@ -1,3 +1,3 @@
 Version 0.7.9
 DataVersion 0
-6.10-7.6
```
bocekm: This is already part of another PR.

bocekm (changes requested):
Otherwise LGTM.

version
pirat89: you are right. it's here by mistake. good catch

bocekm (approved):

PR #129, opened 2020-07-07
others/configchanges: fix syntax error

pirat89:
The issue happens when a file is included in multiple rpms.
E.g.: pam.x86_64,  pam.i686. Thanks Masahiro for the fix.

RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1633894

Co-authored-by: Masahiro Matsuya <mmatsuya@redhat.com>

### DIFF
--- a/RHEL6_7/others/configchanges/check
+++ b/RHEL6_7/others/configchanges/check
@@ -35,10 +35,15 @@ while read line; do
 
   # when a config file is removed by user, rpm -qcV returns string
   # 'missing  c /path/file' instead of usual output
-
-  fflags=$(rpm -q --qf '[%{filenames}: %{fileflags}\n]' -f ${flname} | grep "^${flname}:" | cut -d ":" -f 2)
-  [ $[ $fflags & 16 ] -eq 16 ] && \
-    noreplace_files+=("${flname}")
+  for fflags in $(rpm -q --qf '[%{filenames}: %{fileflags}\n]' -f ${flname} | grep "^${flname}:" | cut -d ":" -f 2)
+    do
+      # Only the 1 << 4 b (== 16) matters here. That is the RPMFILE_NOREPLACE
+      # flag. So mask the other bits.
+      if [ $[ $fflags & 16 ] -eq 16 ]; then
+        noreplace_files+=("${flname}")
+        break
+      fi
+    done
 
 done < $VALUE_CONFIGCHANGED
 


RHEL6_7/others/configchanges/check
```diff
@@ -36,9 +36,13 @@ while read line; do
   # when a config file is removed by user, rpm -qcV returns string
   # 'missing  c /path/file' instead of usual output
 
-  fflags=$(rpm -q --qf '[%{filenames}: %{fileflags}\n]' -f ${flname} | grep "^${flname}:" | cut -d ":" -f 2)
-  [ $[ $fflags & 16 ] -eq 16 ] && \
-    noreplace_files+=("${flname}")
+  for fflags in $(rpm -q --qf '[%{filenames}: %{fileflags}\n]' -f ${flname} | grep "^${flname}:" | cut -d ":" -f 2)
+    do
+      if [ $[ $fflags & 16 ] -eq 16 ]; then
```
bocekm, line 42: Why do you clear first 4 bits (`& 16`)? This condition passes when the flag is `17`. Is it intended? A comment for the choice of this condition would be great.

pirat89: @bocekm  Yes. The only one bit matters in INT32 here (actually I am clearing all bits excluding that one, not only the first four):

| Name              | Value    |
| ----------------- | -------- |
| RPMFILE_CONFIG    | (1 << 0) |
| RPMFILE_DOC       | (1 << 1) |
| RPMFILE_DONOTUSE  | (1 << 2) |
| RPMFILE_MISSINGOK | (1 << 3) |
| RPMFILE_NOREPLACE | (1 << 4) |
| RPMFILE_SPECFILE  | (1 << 5) |
| RPMFILE_GHOST     | (1 << 6) |
| RPMFILE_LICENSE   | (1 << 7) |
| RPMFILE_README    | (1 << 8) |
| RPMFILE_EXCLUDE   | (1 << 9) |

In this case, 17 is importatnt as well, as it refers to config that has noreplace flag. But the RPMFILE_NOREPLACE flag is the only that matters in this module really.

pirat89:
Rebased against upstream branch, to enable testing for the current upgrade path.

RHEL6_7/others/configchanges/check
bocekm: Thanks ;)

bocekm (approved):

PR #133, opened 2020-07-20
 UEFI: fix 'back up' in the pre/post upgrade script

pirat89:
Originally the pre-upgrade script backed up the efi binary file and
    the grub configuration file. But in case the rollback capability is
    used with r-u-t (fixed one; original upstream r-u-t doesn't support
    rollbacks on EFI boot), the backed up version of the file could
    already exist - and another back up during the execution of the
    pre-upgrade script could override the original one. That means
    the rollback could produce a grub configuration that contains
    invalid entries.
    
    As well, I have been thinking about validity of the back up of the
    configuration file by the pre-upgrade script, as in case the
    configuration file is removed during the upgrade, the restore from
    the back up in majority cases will not safe anything as it will
    contain usually invalid entries. But in some cases, it could 'safe
    world'. E.g. manually created entries without referrence to specific
    kernel version & release (e.g. rescue entry) could be still working
    and provide required environment to rescue the machine (or e.g.
    create valid bootloader configuration). So I am keeping it there,
    as in case that bootloader is already broken, it would only 'safe the
    world'.

### DIFF
--- a/RHEL6_7/system/uefi/efibootorder_fix.sh
+++ b/RHEL6_7/system/uefi/efibootorder_fix.sh
@@ -23,11 +23,17 @@ if [ -e "$efibin_path" ] ; then
     # backup these files
     log_info "Backing up EFI files."
     cp -a ${efibin_path}{,.preupg}
-    cp -a ${eficonf_path}{,.preupg}
+    # back up the file only in case there is not any other backup
+    # - the backup can be created by redhat-upgrade-tool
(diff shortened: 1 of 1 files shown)


pirat89:
~@AloisMahdal  Man, I think that this change will need your review ;-)~

Rezney (approved):
Discussed offline. LGTM.

PR #134, opened 2020-07-21
Bump to version 0.8.0

pirat89:
- Switch the upgrade path to RHEL 6.10 -> RHEL 7.9
- Enable upgrades with EFI boot
- Inhibit upgrade when by-path is used in /etc/fstab
- Drop the broken InterVariants module
- Fix and improve migration of tuned
- Fix networking issues around network-manager and route-eth* files
  (RHBZ: #1644915, #1816751)
- Fix report from the networking/openldap module
- Fix syntax error in the other/configchanges module
- Handle dead symlinksi in the system/SysconfigCgroupDaemon module


AloisMahdal (approved):
Bump LGTM.

PR #135, opened 2021-03-08
Uefi fix

pirat89:
In case r-u-t is called without the --cleaup-post option, the postupgrade script recovering the EFI (bin and config) files do not recover the files because the legacy grub is still installed.

In addition, several other minor issues have been present. Fixing those:

- Split the pre & post upgrade script into two separate files to make it less buggy
- Pre-upgrade script:
  - fixed minor issues for the back up of grub files during pre-upgrade checks
  - stop the upgrade if the efi binary file does not exist (it must; this is just seatbelt, it is not expected issue and there is not valid installation of RHEL that could be used as reproducer)
- Post-upgrade script:
  - remove the grub rpm if installed
  - install grub2, grub2-efi, and efiboot rpms if any missing
  - always restore the grub.efi binary file (that's difference between broken bootloader and working one...)
  - restore grub config file if original one dissapeared
 
https://bugzilla.redhat.com/show_bug.cgi?id=1915393

### DIFF
--- a/RHEL6_7/system/uefi/check
+++ b/RHEL6_7/system/uefi/check
@@ -24,10 +24,10 @@ command -v efibootmgr >/dev/null || {
 }
 
 # create pre-upgrade and post-upgrade scripts
-cp -a "efibootorder_fix.sh" "$PREUPGRADE_SCRIPT_DIR/"
-cp -a "efibootorder_fix.sh" "$POSTUPGRADE_DIR/"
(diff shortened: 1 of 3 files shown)


pirat89:
CI is stucked. Creating the build manually.

Rezney:
Looks good to me.

Rezney (approved):

bocekm (approved):

